PR #5, opened 2017-01-12
Migrate Celery -> RQ

davidmcclure:
This replaces Celery with RQ, and makes a couple other small changes to make deployment / debugging easier.

- In `syllascrape.tasks`, refactor the `crawl()` method to work with RQ.
- Provide a Redis connection for RQ in `syllascrape.services`.
- Move `make_params()` into `syllascraps.tasks`, so it can be imported in the REPL and tested with `crawl()`.
- Read configuration settings from `.env` in the root directory, instead of trying to import a local submodule inside `settings`.


wearpants:
looks good

PR #9, opened 2017-01-30
custom spider integration

wearpants:
small cleanup for integrating custom spiders


HarrisonGregg:
Thanks for the refactor.  The custom spiders implementation feels a lot less hacked together now.  Everything looks good, but I'm just going to be nitpicky about one thing.  `extract_links` was described as returning a list of tuples, but all implementations were actually generating tuples.  I changed the docstrings to describe it as a generator, and changed the base implementation back to an empty generator instead of returning an empty tuple for consistency. With that change, looks good to go.

PR #13, opened 2017-03-09
Add Reedley College scraper

jodizzle:
http://reedleycollege.edu/index.aspx?page=903

This is an ASP.Net site with esoteric parameters for POST requests, but thankfully `scrapy.FormRequest.from_response` seems to be able to handle it.

Some folders contain sub-folders, which prompts the if-else recursive logic in `open_folders`.

I'm a little bit concerned that I'm not setting the `meta` content for requests sensibly, so if there's any commentary on that, it would be much appreciated.

### DIFF
--- a/osp_scraper/spiders/reedley.py
+++ b/osp_scraper/spiders/reedley.py
@@ -6,52 +6,44 @@ from ..spiders.CustomSpider import CustomSpider
 
 class ReedleySpider(CustomSpider):
     name = "reedley"
-    allowed_domains = ["reedleycollege.edu"]
 
-    def start_requests(self):
-        start_url = "http://reedleycollege.edu/index.aspx?page=903"
+    start_urls = ["http://reedleycollege.edu/index.aspx?page=903"]
 
-        def open_folders(response):
-            tags = response.css('#_ctl0_listDataGrid tr td:nth-child(2) a')
-            # The hrefs for links that are not folders are documents that
-            # need to be sent through the pipeline.
-            if response.meta['hops_from_seed'] != 0:
-                for item in self.parse_for_files(response):
-                    print('inside', response.meta['source_anchor'], response.meta['depth'])
-                    yield item
+    def parse(self, response):
+        tags = response.css('#_ctl0_listDataGrid tr td:nth-child(2) a')
+        for tag in tags:
+            href = tag.css('::attr(href)')
+            anchor = tag.css('a::text').get()
+            path = response.meta.get("source_anchor", "") + "/" + anchor
 
-            for tag in tags:
+            if href.get().startswith("javascript"):
                 # The hrefs for links that are folders are javascript commands,
                 # so extract the useful bits to be used in recursive POST
                 # requests that open those folders.
-                link = tag.css('::attr(href)').re_first(r'_ctl0.*?_ctl0')
-                if link:
-                    event_target = link.replace('$', ':')
-                    anchor = tag.css('a::text').extract_first()
-                    print(event_target, anchor)
+                folder_link = href.re_first(r'_ctl0.*?_ctl0')
+                event_target = folder_link.replace('$', ':')
 
-                    yield scrapy.FormRequest.from_response(
-                        response,
-                        formdata={
-                            '__EVENTTARGET': event_target,
-                        },
-                        meta={
-                            'depth': 1,
-                            'hops_from_seed': 1,
-                            'source_url': response.url,
-                            'source_anchor': anchor,
-                            'require_files': True
-                        },
-                        method='POST',
-                        callback=open_folders
-                    )
-
-        yield scrapy.Request(start_url, meta={'hops_from_seed': 0}, callback=open_folders)
-
-    def extract_links(self, response):
-        for tag in response.css('#_ctl0_listDataGrid tr td:nth-child(2)'):
-            link = tag.css('a::attr(href)').extract_first()
-            if not 'javascript' in link:
-                anchor = tag.css('a::text').extract_first()
-                url = response.urljoin(link)
-                yield (url, anchor)
+                yield scrapy.FormRequest.from_response(
+                    response,
+                    method='POST',
+                    formdata={
+                        '__EVENTTARGET': event_target,
+                    },
+                    meta={
+                        'depth': response.meta["depth"] + 1,
+                        'hops_from_seed': response.meta["hops_from_seed"] + 1,
+                        'source_anchor': path
+                    },
+                    callback=self.parse
+                )
+            else:
+                yield response.follow(
+                    href.get(),
+                    meta={
+                        'depth': response.meta["depth"] + 1,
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/reedley.py
```diff
@@ -0,0 +1,68 @@
...
+                # The hrefs for links that are folders are javascript commands,
+                # so extract the useful bits to be used in recursive POST
+                # requests that open those folders.
+                if 'javascript' in link:
+                    prefix = '_ctl0$listDataGrid$_ctl'
+                    suffix = '$_ctl0'
+                    # There's a number in between prefix and suffix that differs
+                    # among folder links.
+                    start = link.find(prefix)
+                    end = link.find(suffix, start) + len(suffix)
```
HarrisonGregg: Doing it this way might be a little more robust, but looking at webpage, start should always be 25 and end should always equal -5, so I'm not sure it's worth the added complexity.

osp_scraper/spiders/reedley.py
```diff
@@ -0,0 +1,68 @@
...
+                else:
+                    for item in self.parse_for_files(response):
+                        yield item
+
+        yield scrapy.Request(
+            start_url,
+            meta={
+                'depth': 0,
+                'hops_from_seed': 0,
+                'source_url': "",
```
HarrisonGregg: There's no need to include these here as empty strings.

jodizzle: Yeah.  Not sure why I included a `meta` dict here anyway.

HarrisonGregg:
@jodizzle With these two changes, should be good to go.

osp_scraper/spiders/reedley.py
jodizzle: Is it really that much more complex?  I was under the impression that this was (sort of) a python idiom.

Also, another possibility is that I could split by single quotation marks, e.g.,

```event_target = link.split("'")[1].replace('$', ':')```

But I'm not sure how clear or safe that is.

Comment (jodizzle) [inline osp_scraper/spiders/reedley.py]: Oh I didn't know about `re_first`.  I can try it out.

File: osp_scraper/spiders/reedley.py [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+                            'source_anchor': anchor,
+                        },
+                        method='POST',
+                        callback=open_folders
+                    )
+
+                # The hrefs for links that are not folders are documents that
+                # need to be sent through the pipeline.
+                else:
+                    for item in self.parse_for_files(response):
```
jodizzle: Oh oops, I didn't notice that.  That's especially bad because I think all the pages on the site will either be all folders or all links to syllabi, no mix, so I guess this `else` statement is triggered once for each syllabi on the syllabi pages.  I'll need to change the scraper's behavior around to avoid this.

osp_scraper/spiders/reedley.py
```diff
@@ -0,0 +1,56 @@
...
+                            'source_url': response.url,
+                            'source_anchor': anchor,
+                        },
+                        method='POST',
+                        callback=open_folders
+                    )
+
+                # The hrefs for links that are not folders are documents that
+                # need to be sent through the pipeline.
+                elif not link and not is_parsed:
```
HarrisonGregg: You don't have to change anything, but in general, I wouldn't be too worried about yielding a few extra PageItems, given that these files will be treated the same as the files from the general scraper that grabs a ton of extraneous files.

osp_scraper/spiders/reedley.py
```diff
@@ -0,0 +1,56 @@
...
+                    anchor = tag.css('a::text').extract_first()
+
+                    yield scrapy.FormRequest.from_response(
+                        response,
+                        formdata={
+                            '__EVENTTARGET': event_target,
+                        },
+                        meta={
+                            'depth': 1,
+                            'hops_from_seed': 1,
```
HarrisonGregg: Given that this one is recursive, shouldn't it be incrementing these every time?

osp_scraper/spiders/reedley.py
```diff
@@ -0,0 +1,56 @@
...
+                            'source_url': response.url,
+                            'source_anchor': anchor,
+                        },
+                        method='POST',
```
HarrisonGregg: Using `from_response`, is specifying the method actually necessary, or is this just for clarity?

osp_scraper/spiders/reedley.py
```diff
@@ -0,0 +1,56 @@
...
+class ReedleySpider(CustomSpider):
+    name = "reedley"
+    allowed_domains = ["reedleycollege.edu"]
+
+    def start_requests(self):
+        start_url = "http://reedleycollege.edu/index.aspx?page=903"
+
+        def open_folders(response):
+            tags = response.css('#_ctl0_listDataGrid tr td:nth-child(2) a')
+            is_parsed = False
```
HarrisonGregg: Again, you don't have to change anything, but personally, the way I would do this check is with a `has_files` variable, and then put the `parse_for_files` loop outside the tags loop to emphasize that it's only executed once or not at all.

osp_scraper/spiders/reedley.py
jodizzle: Just for clarity.

osp_scraper/spiders/reedley.py
jodizzle: Ah, good point.  I missed that.

jodizzle:
@HarrisonGregg I realized recently while looking at some printouts for this scraper that there's an issue where the `POST` requests made inside the folders that have subfolders (e.g., '1999 Fall Syllabi') are very similar to `POST` requests made to top-level folders.  This seems to lead Scrapy to skip over a whole bunch of the top-level folders (or possibly, I suppose, delay them; I only ran each test for a small amount of time).  Seems like this could be another iteration of #69.

If you want to take a look you're welcome to, but I'm going to leave this PR open for now and return to it later.

HarrisonGregg:
I think the potential problem you're talking about is that `__EVENTTARGET` is the same for some of the root level and nested folders.  Looking into it, I don't think this is actually a problem.  I think that `VIEWSTATE` actually contains information about what the links on the page are, so it can actually differentiate between the requests on the different pages.

jodizzle:
@HarrisonGregg Hmm, It's possible I was wrong, but I remember testing it and running into problems.  I'll look into it again at some point soon.

HarrisonGregg:
If you want to test it, I commented out the part that downloads the files and ran it to check that it made it to all of the folders, which it seemed to just fine.

jodizzle:
@HarrisonGregg Alright, I didn't test super closely, but everything seems to be in order.  Not sure what the problem was before.  It may have just been my imagination and/or faulty testing.

I really like how you cleaned it up, thanks for that.

jodizzle:
OK to merge (unfortunately, I can't give a Github-official stamp of approval to my own PR).

PR #14, opened 2017-03-20
Add Clemson University scraper

jodizzle:
https://etpr.app.clemson.edu/repository/syllabus_public.php

Funny example of a "captcha", but otherwise a straightforward scraper.

### DIFF
--- a/osp_scraper/spiders/clemson.py
+++ b/osp_scraper/spiders/clemson.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+
+import itertools
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class ClemsonSpider(CustomSpider):
+    name = "clemson"
+
+    def start_requests(self):
+        start_url = "https://etpr.app.clemson.edu/repository/syllabus_public.php"
+        # Need to answer a captcha.
+        # Captcha is always the same (pseudo-captcha).
+        yield scrapy.FormRequest(
+            start_url,
+            formdata={
+                'person_ck': 'CLEMSON',
+                'check_person': 'Logon',
+            },
+            method='POST',
+            callback=self.parse_for_search_terms
+        )
+
+    def parse_for_search_terms(self, response):
+        sem_codes = response.css('select[name="semester_selected"] option')\
+                            .css('::attr(value)')\
+                            .extract()
+        # Skip the first entry because it's an empty string.
+        disciplines = response.css('select[name="subj_selected"] option')[1:]
+        subj_codes = disciplines.css('option::attr(value)').extract()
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/clemson.py
```diff
@@ -0,0 +1,68 @@
...
+                    'hops_from_seed': 1,
+                    'source_url': response.url,
+                    'source_anchor': '',
+                },
+                callback=get_searches
+            )
+
+
+        def get_searches(response):
+            semesters = response.css('select[name="semester_selected"] option')
```
HarrisonGregg: I like using `extract_first` and then checking if it was successful, but if you're not going to, you could shorten your code by using `extract` with something like: `response.css('select[name="semester_selected"] option::attr(value)').extract()`.

osp_scraper/spiders/clemson.py
```diff
@@ -0,0 +1,68 @@
...
+            # Skip the first entry because it's an empty string.
+            disciplines = response.css('select[name="subj_selected"] option')[1:]
+            for sem, subj in itertools.product(semesters, disciplines):
```
HarrisonGregg: Out of curiosity, why use `itertools.product` instead of just a nested for loop?

jodizzle:
@HarrisonGregg I updated the looping.  It elongates the process a little bit, but I think it's still readable.

osp_scraper/spiders/clemson.py
jodizzle: I just feel like it's an appropriate use of Python's libraries.  Especially helps with keeping the number of indentation levels in check.

osp_scraper/spiders/clemson.py
```diff
@@ -0,0 +1,67 @@
...
+from ..spiders.CustomSpider import CustomSpider
+
+class ClemsonSpider(CustomSpider):
+    name = "clemson"
+    allowed_domains = ["clemson.edu"]
+
+    def start_requests(self):
+        start_url = "https://etpr.app.clemson.edu/repository/syllabus_public.php"
+
+        def do_captcha(response):
```
jodizzle: Ah, yes, that makes more sense.

osp_scraper/spiders/clemson.py
```diff
@@ -0,0 +1,67 @@
...
+                    'source_url': response.url,
+                    'source_anchor': '',
+                },
+                callback=get_searches
+            )
+
+
+        def get_searches(response):
+            semesters = response.css('select[name="semester_selected"] option')
+            sem_codes = semesters.css('option::attr(value)').extract()
```
jodizzle: Sure, we can try that.

jodizzle:
@HarrisonGregg I refactored.

PR #16, opened 2017-03-20
Add Cedar Crest College scraper

jodizzle:
https://my.cedarcrest.edu/ics/staticpages/syllabilist.aspx

Single option form.  Straightforward.

There were some 404's when trying to download files during testing, but I believe those are from actual missing files on the server and not errors in the requests.

### DIFF
--- a/osp_scraper/spiders/cedarcrest.py
+++ b/osp_scraper/spiders/cedarcrest.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class CedarCrestSpider(CustomSpider):
+    name = "cedarcrest"
+    allowed_domains = ["cedarcrest.edu"]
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/cedarcrest.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class CedarCrestSpider(CustomSpider):
+    name = "cedarcrest"
+    allowed_domains = ["cedarcrest.edu"]
```
HarrisonGregg: It doesn't matter, but I've started leaving out `allowed_domains`, since we know that custom spiders won't get carried away and there might be files linked on a different domain sometimes.

jodizzle: That's a good point, but I think I'll keep it for now.

PR #18, opened 2017-03-20
Add Woodbury University scraper

jodizzle:
https://go.woodbury.edu/syllabus/view.aspx

Two option form.  Straightforward.

### DIFF
--- a/osp_scraper/spiders/woodbury.py
+++ b/osp_scraper/spiders/woodbury.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+
+import itertools
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class WoodburySpider(CustomSpider):
+    name = "woodbury"
+    allowed_domains = ["woodbury.edu"]
+
+    def start_requests(self):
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/woodbury.py
```diff
@@ -0,0 +1,48 @@
...
+    allowed_domains = ["woodbury.edu"]
+
+    def start_requests(self):
+        start_url = "https://go.woodbury.edu/syllabus/view.aspx"
+
+        def get_years(response):
+            years = response.css('select#dYear option')
+            terms = response.css('select#dTerm option')
+            for year, term in itertools.product(years, terms):
+                dYear = year.css('option::attr(value)').extract_first()
```
HarrisonGregg: I'm not sure how much of a difference it would make practically, but on principle, I would avoid parsing the HTML over and over again like this.

jodizzle: I imagine it's a pretty small factor, but you're right.  I wasn't considering how I was extracting some of the same data over and over again.

jodizzle:
@HarrisonGregg I changed the loop in `get_years` based on your suggestion.  I also realized that `option::attr(value)` and `option::attr(text)` are always the same for this site, so that helps simplify things.

PR #19, opened 2017-03-20
Add Sam Houston State University scraper

jodizzle:
https://ww2.shsu.edu/faci10wp/

Two part, three option form.  Being able to choose "All Departments" as an option is convenient, however it seems to take a long time to fetch the resulting document (one page with many courses on it).  Is this an issue?  I could instead do one request per department.

### DIFF
--- a/osp_scraper/spiders/shsu.py
+++ b/osp_scraper/spiders/shsu.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+
+import itertools
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SHSUSpider(CustomSpider):
+    name = "shsu"
+    allowed_domains = ["shsu.edu"]
+
+    def start_requests(self):
+        start_url = 'https://ww2.shsu.edu/faci10wp/'
+
+        def get_terms(response):
+            years = response.css('select[name="year"] option::attr(value)')
+            semesters = response.css('select[name="semester"] option::attr(value)')
+            searches = itertools.product(years.extract(), semesters.extract())
+            for year, semester in searches:
+                yield scrapy.FormRequest(
+                    response.urljoin('display.php'),
+                    formdata={
+                        'semester': semester,
+                        'year': year,
+                        # '-' means search all departments.
+                        'dept': '-',
+                    },
+                    method='POST',
+                    meta={
+                        'depth': 1,
+                        'hops_from_seed': 1,
+                        'source_url': response.url,
+                        'source_anchor': year + ' ' + semester,
+                    },
+                    callback=self.parse_for_files
+                )
+
+        yield scrapy.Request(start_url, callback=get_terms)
+
+    def extract_links(self, response):
+        tags_root = response.css('tr.noshade td:nth-child(2)')
+        # Each cell contains multiple course sections that need to be selected.
+        syllabus_links = tags_root.css('font ~ span')
+        class_numbers = tags_root.css('font')
+        for syllabus, number in zip(syllabus_links, class_numbers):
+            link = syllabus.css('a::attr(href)').extract_first()
+            # Skip iteration if the course doesn't have a syllabus link.
+            if not link:
+                continue
+            url = response.urljoin(link)
+            number_text = number.css('font::text').extract_first()
+            # The anchor contains lots of CRLF line endings, so clean it up.
+            anchor = ' '.join(number_text.split())
+            yield (url, anchor)


osp_scraper/spiders/shsu.py
```diff
@@ -0,0 +1,56 @@
...
+    allowed_domains = ["shsu.edu"]
+
+    def start_requests(self):
+        start_url = 'https://ww2.shsu.edu/faci10wp/'
+
+        def get_terms(response):
+            years = response.css('select[name="year"] option')
+            semesters = response.css('select[name="semester"] option')
+            for year, semester in itertools.product(years, semesters):
+                year_data = year.css('option::attr(value)').extract_first()
```
HarrisonGregg: Could you refactor this to extract each year and semester only once instead of over and over again?

jodizzle: Good catch.

jodizzle:
@HarrisonGregg

osp_scraper/spiders/shsu.py
```diff
@@ -0,0 +1,55 @@
...
+                        'source_url': response.url,
+                        'source_anchor': year + ' ' + semester,
+                    },
+                    callback=self.parse_for_files
+                )
+
+        yield scrapy.Request(start_url, callback=get_terms)
+
+    def extract_links(self, response):
+        tags_root = response.css('tr.noshade td:nth-child(2)')
```
HarrisonGregg: You might want to include a comment here saying you can't iterate through the rows or cells because each contains multiple sections of a course.

osp_scraper/spiders/shsu.py
```diff
@@ -0,0 +1,55 @@
...
+            years = response.css('select[name="year"] option::attr(value)')
+            semesters = response.css('select[name="semester"] option::attr(value)')
+            searches = itertools.product(years.extract(), semesters.extract())
+            for year, semester in searches:
+                yield scrapy.FormRequest(
+                    response.urljoin('display.php'),
+                    formdata={
+                        'semester': semester,
+                        'year': year,
+                        # Using '-' here signifies all departments being
```
HarrisonGregg: As below, your comments could be more concise sometimes.  Maybe just:
```python
# '-' means search all departments
```

osp_scraper/spiders/shsu.py
```diff
@@ -0,0 +1,55 @@
...
+
+        yield scrapy.Request(start_url, callback=get_terms)
+
+    def extract_links(self, response):
+        tags_root = response.css('tr.noshade td:nth-child(2)')
+        syllabus_links = tags_root.css('font ~ span')
+        class_numbers = tags_root.css('font')
+        for syllabus, number in zip(syllabus_links, class_numbers):
+            link = syllabus.css('a::attr(href)').extract_first()
+            # Some span blocks do not actually contain 'a' tags because the
```
HarrisonGregg: Some of your comments, like this one, could be a little more concise.  Not all courses having syllabi is a common pattern, so if you must include a comment, it could fit on one line.  Something like:
```python
# Skip iteration if course doesn't have syllabus link
```

osp_scraper/spiders/shsu.py
```diff
@@ -0,0 +1,55 @@
...
+from ..spiders.CustomSpider import CustomSpider
+
+class SHSUSpider(CustomSpider):
+    name = "shsu"
+    allowed_domains = ["shsu.edu"]
+
+    def start_requests(self):
+        start_url = 'https://ww2.shsu.edu/faci10wp/'
+
+        def get_terms(response):
```
HarrisonGregg: Given that you care more about line length, you could opt for implementing `parse` or using a non-nested function to save a few spaces.

osp_scraper/spiders/shsu.py
jodizzle: Agreed.

osp_scraper/spiders/shsu.py
jodizzle: Agreed.

osp_scraper/spiders/shsu.py
jodizzle: Ah, I think you're right, that probably does warrant some more explanation.

osp_scraper/spiders/shsu.py
jodizzle: I actually like the style of nesting functions inside `start_requests`, so I think I'll keep doing that.

How would using `parse` help with line length?

jodizzle: Actually, I think I should adjust the style of this scraper a bit, but I'll save it for later whenever we do a batch set of changes.

PR #20, opened 2017-03-20
Allow file pipeline to follow redirects

HarrisonGregg:
@wearpants This is the _hack_ that I mentioned for getting the file pipeline to follow redirects instead of just giving a warning and not downloading the file.  I couldn't figure out how to do this for individual spiders, so we should verify somehow that it won't break anything else.

### DIFF
--- a/osp_scraper/pipelines.py
+++ b/osp_scraper/pipelines.py
@@ -254,6 +254,9 @@ class WarcFilesPipeline(FilesPipeline):
         return path_from_warc(record, info.spider.run_id)
 
     def _check_media_to_download(self, result, request, info):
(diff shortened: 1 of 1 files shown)


osp_scraper/pipelines.py
```diff
@@ -252,3 +252,8 @@ def file_path(self, request, response=None, info=None):
             record = response.meta['warc_record']
 
         return path_from_warc(record, info.spider.run_id)
+
+    def _check_media_to_download(self, result, request, info):
+        x = super()._check_media_to_download(result, request, info)
```
wearpants: please add a comment about what this does & why it is here

PR #26, opened 2017-03-20
Scraper/campusconcourse

HarrisonGregg:


jodizzle:
I'm a bit confused by this history.  Rebase?

HarrisonGregg:
@jodizzle How about now?

osp_scraper/spiders/campusconcourse.py
```diff
@@ -0,0 +1,92 @@
...
+                yield scrapy.FormRequest(
+                    response.url, 
+                    method='GET',
+                    formdata={
+                        "search_performed": "1",
+                        "offset": str(page)
+                    },
+                    meta={
+                        'depth': 1,
+                        'hops_from_seed': 1,
```
HarrisonGregg: Maybe it would be different if you were clicking through manually, but here, all of the requests are one step away from the initial request, so `depth`/`hops_from_seed` of `1` seems correct.

osp_scraper/spiders/campusconcourse.py
```diff
@@ -0,0 +1,92 @@
...
+                    },
+                    callback=self.parse_results_page
+                )
+
+        for start_url in self.start_urls:
+            yield scrapy.FormRequest(
```
HarrisonGregg: I'm not sure exactly what you're proposing, but the initial requests do need to have `search_performed=1` added, so it can't be done with out implementing `start_requests`.

The `edu_repo_crawler` sets `start_urls` to the URLs in the `Database URLs` column.

HarrisonGregg:
@jodizzle I've refactored significantly to move the functionality for downloading files to a [separate scraper](https://github.com/opensyllabus/osp-scraper/pull/53).

osp_scraper/spiders/campusconcourse.py
jodizzle: All I was saying is that `parse` seems at first like it would be a good idea because it gets handed responses from the `start_urls` automatically and looks cleaner, but it actually isn't, because you need to add extra `formdata`.  So what you said, basically.

jodizzle:
@HarrisonGregg I think in some ways it would be nicer to have both functionalities combined in one scraper, but I think what you've done here is fine.

PR #27, opened 2017-03-20
Scraper/lakelandcollege

HarrisonGregg:


osp_scraper/spiders/lakelandcollege.py
```diff
@@ -0,0 +1,58 @@
...
+                )
+
+        yield scrapy.Request(terms_url, callback=get_terms)
+
+    def extract_links(self, response):
+        # The xpath is necessary here because the response body has two <html> tags
+        relative_url = response.xpath('//a').css('a::attr(href)').extract_first()
+        if relative_url:
+            url = response.urljoin(relative_url)
+            anchor = response.css('#description_container a::text').extract_first()
```
jodizzle: I don't think this anchor works.  Seems to lead to a `X-Source-Anchor: None`.

osp_scraper/spiders/lakelandcollege.py
```diff
@@ -0,0 +1,58 @@
...
+                        'hops_from_seed': 2,
+                    },
+                    callback=self.parse_for_files
+                )
+
+        yield scrapy.Request(terms_url, callback=get_terms)
+
+    def extract_links(self, response):
+        # The xpath is necessary here because the response body has two <html> tags
+        relative_url = response.xpath('//a').css('a::attr(href)').extract_first()
```
jodizzle: This is true when accessing the site normally, because the course window appears as a popup window, but the response received when using `scrapy` is the popup window as a distinct page with only a single pair of `html` tags.  So you can probably stick with regular CSS selectors.

HarrisonGregg: I meant that the response is actually malformed HTML.  You can see it in the actual response to a GET request to this URL: [https://www.lakelandcollege.edu/as/as/academicprograms/catalog/includes/popup/course_descrip.cfm?crs_name=CIS-160&crs_title=Practical%20Software%20Application&id=14406](https://www.lakelandcollege.edu/as/as/academicprograms/catalog/includes/popup/course_descrip.cfm?crs_name=CIS-160&crs_title=Practical%20Software%20Application&id=14406)

HarrisonGregg:
@jodizzle I think I made all the changes. How does it look now?

osp_scraper/spiders/lakelandcollege.py
jodizzle: Oh, weird.  I guess using `xpath` makes sense then.

osp_scraper/spiders/lakelandcollege.py
```diff
@@ -0,0 +1,60 @@
...
+                url = response.urljoin(relative_url)
+                yield scrapy.Request(url, callback=get_courses)
+
+        def get_courses(response):
+            for a_tag in response.css("table a"):
+                relative_url = a_tag.css("a::attr(href)").extract_first()
+                url = response.urljoin(relative_url)
+                anchor = a_tag.css("a::text").extract_first()
+                yield scrapy.Request(
+                    url,
```
jodizzle: Minor: Trailing whitespace.

osp_scraper/spiders/lakelandcollege.py
```diff
@@ -0,0 +1,60 @@
...
+        def get_programs(response):
+            for a_tag in response.css("table a"):
+                relative_url = a_tag.css("a::attr(href)").extract_first()
+                url = response.urljoin(relative_url)
+                yield scrapy.Request(url, callback=get_courses)
+
+        def get_courses(response):
+            for a_tag in response.css("table a"):
+                relative_url = a_tag.css("a::attr(href)").extract_first()
+                url = response.urljoin(relative_url)
```
jodizzle: There's at least one case where grabbing `url`s like this will badly error: https://www.lakelandcollege.edu/as/as/academicprograms/catalog/list.cfm?acadprog=CRT.PN.TRK&catalog_term=18SP.  Note that there are several `a` tags in a table, and most of them are PDF files, which will lead to handing `extract_links` binary response data.

The trait that the pop-up window `a` tags seem to have in common is the `fancyiframe1` class, so you should probably select for that.  There might be other cases of other types of links you would want to select, but if the course pages are really varied it's probably not worth focusing on too hard.

osp_scraper/spiders/lakelandcollege.py
```diff
@@ -0,0 +1,60 @@
...
+                anchor = a_tag.css("a::text").extract_first()
+                yield scrapy.Request(
+                    url, 
+                    meta={
+                        'source_url': response.url,
+                        'source_anchor': anchor,
+                        'depth': 0,
+                        'hops_from_seed': 3
```
HarrisonGregg: Given that depth gets reset to 0 on every request, I'm not sure it makes sense to set `depth` to `0`.

HarrisonGregg:
@jodizzle

PR #28, opened 2017-03-20
Scraper/utrgv

HarrisonGregg:


osp_scraper/spiders/utrgv.py
```diff
@@ -0,0 +1,49 @@
...
+                )
+
+        def parse_results_page(response):
+            next_relative_url = response.css(".table-pagination a::attr(href)").extract_first()
+            if next_relative_url:
+                url = response.urljoin(next_relative_url)
+                yield scrapy.Request(url, callback=parse_results_page)
+
+            for rq in self.parse_for_files(response):
+                yield rq
```
jodizzle: Why is the variable name `rq`?

osp_scraper/spiders/utrgv.py
```diff
@@ -0,0 +1,49 @@
...
+
+        yield scrapy.Request(start_url, callback=get_term_codes)
+
+    def extract_links(self, response):
+        for row in response.css('.course-listing tr'):
+            relative_url = row.css('.syllabus-icon a::attr(href)').extract_first()
+            if relative_url:
+                url = response.urljoin(relative_url)
+                anchor = row.css("tr::text").extract_first()
```
jodizzle: Leads to a blank anchor.  You might want to extract from the `td` elements instead.

osp_scraper/spiders/utrgv.py
```diff
@@ -0,0 +1,49 @@
...
+                        'hops_from_seed': 1,
+                    },
+                    callback=parse_results_page
+                )
+
+        def parse_results_page(response):
+            next_relative_url = response.css(".table-pagination a::attr(href)").extract_first()
+            if next_relative_url:
+                url = response.urljoin(next_relative_url)
+                yield scrapy.Request(url, callback=parse_results_page)
```
jodizzle: I think you should include a `meta` argument here as well.

osp_scraper/spiders/utrgv.py
```diff
@@ -0,0 +1,49 @@
...
+
+        def get_term_codes(response):
+            for option in response.css("#semester-select option"):
+                term_code = option.css("option::attr(value)").extract_first()
+                name = option.css("option::text").extract_first()
+
+                yield scrapy.FormRequest(
+                    start_url, 
+                    formdata={
+                        "semester":term_code
```
jodizzle: Minor: Space after colon.

HarrisonGregg:
@jodizzle I think I've fixed everything. How does it look now?

osp_scraper/spiders/utrgv.py
```diff
@@ -0,0 +1,57 @@
...
+                    'hops_from_seed': response.meta['hops_from_seed'] + 1
+                },
+                callback=self.parse_results_page
+            )
+
+        for item in self.parse_for_files(response):
+            yield item
+
+    def extract_links(self, response):
+        for row in response.css('.course-listing tr'):
```
jodizzle: Even though I guess your `if` statement below covers it, the first row is the table header, so you can skip it with `[1:]` or rule it out with `tbody`:  `response.css('.course-listing tbody tr')`.

osp_scraper/spiders/utrgv.py
```diff
@@ -0,0 +1,57 @@
...
+            yield scrapy.FormRequest(
+                response.url,
+                formdata={
+                    "semester": term_code
+                },
+                meta={
+                    'source_url': response.url,
+                    'source_anchor': name,
+                    'depth': 0,
+                    'hops_from_seed': 1,
```
jodizzle: Why is `hops_from_seed` 1 here, as opposed to 0?

osp_scraper/spiders/utrgv.py
```diff
@@ -0,0 +1,57 @@
...
+        for item in self.parse_for_files(response):
+            yield item
+
+    def extract_links(self, response):
+        for row in response.css('.course-listing tr'):
+            relative_url = row.css('.syllabus-icon a::attr(href)')\
+                .extract_first()
+            if relative_url:
+                url = response.urljoin(relative_url)
+                anchor = " ".join(row.css('td::text').extract()[:5])
```
jodizzle: I saw a few anchors that had a lot of trailing whitespace, coming from the "Course Title" cell (see the result [here](https://webapps.utrgv.edu/aa/dm/index.cfm?dm_name=QUMT&thecollege=SELECT&thesubject=SELECT&semester=201327)), so it might be a good idea to use a `.split()` as well or call `self.clean_whitespace`.

osp_scraper/spiders/utrgv.py
HarrisonGregg: If the first request, which has `parse` as a callback, has `hops_from_seed` of 0, then it seemed like the next request chained from that one would increment that by 1.  But again, I don't really know how these work.

HarrisonGregg:
@jodizzle Oh, I see what you mean. I didn't notice that some of the PDFs weren't syllabi. I don't see a way to identify them from the results page, and the error page still returns a 200 status, so I'm not sure what we could do to avoid it at the scraper level.  Any ideas?

jodizzle:
@HarrisonGregg I can imagine a couple elaborate things you could do – like elongating the callback process across a few more functions, visiting pages multiple times to check them, carrying around multiple response objects, that sort of thing – but those are probably more trouble than it's worth and would almost certainly end up being inefficient.  I think the best thing to do is probably leave an inline `# NOTE:` about it somewhere in the scraper and then maybe we can address it as a structural edge case later on.

PR #29, opened 2017-03-20
Added functional SFU scraper

HarrisonGregg:


osp_scraper/spiders/sfu.py
```diff
@@ -0,0 +1,55 @@
...
+                url = response.url+"&dept="+department["value"]
+                yield scrapy.Request(url, callback=get_courses)
+
+        def get_courses(response):
+            for course in json.loads(response.body):
+                url = response.url+"&number="+course["value"]
+                yield scrapy.Request(url, meta=course, callback=get_sections)
+
+        def get_sections(response):
+            base_url = start_url + "?" + re.sub(r"[^&]*=","", response.url).replace("&","/")
```
jodizzle: Could you add a comment describing what this line does?

osp_scraper/spiders/sfu.py
```diff
@@ -0,0 +1,55 @@
...
+
+        def get_departments(response):
+            for department in json.loads(response.body):
+                url = response.url+"&dept="+department["value"]
+                yield scrapy.Request(url, callback=get_courses)
+
+        def get_courses(response):
+            for course in json.loads(response.body):
+                url = response.url+"&number="+course["value"]
+                yield scrapy.Request(url, meta=course, callback=get_sections)
```
jodizzle: I think I get what's going on here: you're loading the response body, which is JSON where each member represents the data of a `course`.  Then you're passing each `course` as metadata to `get_sections`, where you extract the `title` to get the `source_anchor`.  Maybe it's just me, but it took me a while to figure out what was going on here, mostly I think because I only found out about the existence of `title` a method later.  Do you think you could comment about the contents of `course` and/or be a bit more drawn and explicit in how you approach this?

osp_scraper/spiders/sfu.py
```diff
@@ -0,0 +1,55 @@
...
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SFUSpider(CustomSpider):
+    name = "sfu"
+
+    def start_requests(self):
+        start_url = "http://www.sfu.ca/outlines.html"
+        search_url = "http://www.sfu.ca/bin/wcm/course-outlines"
+        outline_url = "http://www.sfu.ca/outlines.html"
```
HarrisonGregg: Removed.

osp_scraper/spiders/sfu.py
```diff
@@ -0,0 +1,55 @@
...
+
+        def get_years(response):
+            years = response.css("#year option::attr(value)").extract()
+
+            for year in years[:1]:
```
HarrisonGregg: Yep.

osp_scraper/spiders/sfu.py
HarrisonGregg: I changed it to pass just the title as 'anchor' in meta, which should be more clear.

osp_scraper/spiders/sfu.py
HarrisonGregg: Done.

osp_scraper/spiders/sfu.py
```diff
@@ -0,0 +1,55 @@
...
+                url = response.url+"&number="+course["value"]
+                yield scrapy.Request(url, meta=course, callback=get_sections)
+
+        def get_sections(response):
+            base_url = start_url + "?" + re.sub(r"[^&]*=","", response.url).replace("&","/")
+
+            for section in json.loads(response.body):
+                url = base_url + "/" + section["value"]
+                yield scrapy.Request(
+                    url,
```
HarrisonGregg: Removed.

HarrisonGregg:
@jodizzle

osp_scraper/spiders/sfu.py
```diff
@@ -0,0 +1,61 @@
...
+    name = "sfu"
+
+    start_urls = ["http://www.sfu.ca/outlines.html"]
+
+    def parse(self, response):
+        search_url = "http://www.sfu.ca/bin/wcm/course-outlines"
+        years = response.css("#year option::attr(value)").extract()
+
+        for year in years:
+            url = search_url + "?year=" + year
```
jodizzle: Is there a reason you don't take a string formatting approach here (e.g., `search_url.format(year)` with a slightly modified `search_url`)?

HarrisonGregg: The only reason, which might not be the best, was for consistency with the URL building through the rest of the functions

jodizzle: That's what I suspected.  I think that makes a fair amount of sense.

PR #31, opened 2017-03-20
Scraper/tamu

HarrisonGregg:


osp_scraper/spiders/tamu.py
```diff
@@ -0,0 +1,64 @@
...
+                )
+
+        def select_subject(response):
+            for option in response.css("#subj_id option"):
+                subject_value = option.css("option::attr(value)").extract_first()
+                subject_text = option.css("option::text").extract_first()
+
+                yield scrapy.FormRequest(
+                    search_url,
+                    body=search_body_format.format(response.meta["term_value"], subject_value),
```
jodizzle: Is there a reason you took the approach of formatting a URL and sending it as a request body rather than sending `formdata` as usual?

osp_scraper/spiders/tamu.py
```diff
@@ -0,0 +1,64 @@
...
+
+from ..spiders.CustomSpider import CustomSpider
+
+class TAMUSpider(CustomSpider):
+    name = "tamu"
+
+    def start_requests(self):
+        terms_url =    'https://compass-ssb.tamu.edu/pls/PROD/bwckschd.p_disp_dyn_sched'
+        subjects_url = 'https://compass-ssb.tamu.edu/pls/PROD/bwckgens.p_proc_term_date'
+        search_url =   'https://compass-ssb.tamu.edu/pls/PROD/bwckschd.p_get_crse_unsec'
```
HarrisonGregg: Removed.

osp_scraper/spiders/tamu.py
```diff
@@ -0,0 +1,64 @@
...
+            for option in response.css("#term_input_id option")[1:4]:
+                term_value = option.css("option::attr(value)").extract_first()
+                term_text = option.css("option::text").extract_first()
+                yield scrapy.FormRequest(
+                    subjects_url,
+                    formdata={
+                        "p_calling_proc": "bwckschd.p_disp_dyn_sched",
+                        "p_term": term_value
+                    },
+                    method='POST',
```
HarrisonGregg: Fixed.

osp_scraper/spiders/tamu.py
HarrisonGregg: For some reason, the request gets sent with two copies of some of the parameters with different values, which I couldn't do with `formdata`.

osp_scraper/spiders/tamu.py
```diff
@@ -0,0 +1,64 @@
...
+                    callback=self.parse_for_files
+                )
+
+        yield scrapy.Request(terms_url, callback=select_term)
+
+    def extract_links(self, response):
+        urls = set()
+        for tag in response.css('.datadisplaytable .ddtitle'):
+            url = tag.css('th > span > a::attr(href)').extract_first()
+            if url and url not in urls:
```
HarrisonGregg: I feel like there has to be a reason that I put that check in there, but I can't remember what it could be.  The site doesn't even look like it has duplicate urls on one results page.

osp_scraper/spiders/tamu.py
```diff
@@ -0,0 +1,72 @@
...
+                        "term_value": term_value,
+                        "term_text": term_text
+                    },
+                    callback=select_subject
+                )
+
+        def select_subject(response):
+            for option in response.css("#subj_id option"):
+                subject_value = option.css("::attr(value)").extract_first()
+                subject_text = option.css("::text").extract_first()
```
jodizzle: Oh, I was under the impression that you had to include something before the `::`, in this case something like `option.css("option::text").extract_first()`.  I guess it's pretty neat that you don't have to.

HarrisonGregg:
@jodizzle Good call.  Using lists in `formdata` cleans it up a lot.  I did try submitting it with multiple subjects selected, but the request hadn't finished after a while, so I figured it was safer to just do them one at a time.

osp_scraper/spiders/tamu.py
```diff
@@ -0,0 +1,63 @@
...
+                    'term_value': term_value,
+                    'term_text': term_text
+                },
+                callback=self.parse_for_subjects
+            )
+
+    def parse_for_subjects(self, response):
+        for option in response.css("#subj_id option"):
+            subject_value = option.css("::attr(value)").extract_first()
+            subject_text = option.css("::text").extract_first()
```
jodizzle: So down here, you use `::` at the start of each sub-CSS selector, but above in `parse` you use a more explicit `option::` at the start.  Is one of these preferred?

jodizzle:
True, the time it takes to make a request is important.  I suppose that's something to take into consideration when working on scrapers for sites like these.

PR #33, opened 2017-03-20
Added functional Central Michigan University spider

HarrisonGregg:


HarrisonGregg:
All of the file URLs are generated from the first request made to that URL, so why wouldn't `X-Source-Url` be the same for all of them?

It does seem as though a lot of the links to the syllabi redirect to that not found page, but that's true on the search page itself, so I couldn't figure out any way to detect which ones were missing before requesting them.   It is unfortunate to be making so many unnecessary requests, but it shouldn't affect the output.

jodizzle:
Ah, you're right about the `X-Source-Url`.  I was totally confused.  My bad.

I thought about it, and unfortunately I don't have very many good ideas about how to handle the number of not found pages either.  I suggest we just take note of the issue and leave it for now, since as you said it doesn't affect the output.

jodizzle:
OK to merge.

PR #34, opened 2017-03-20
Added functional DeAnza spider

HarrisonGregg:


osp_scraper/spiders/deanza.py
```diff
@@ -0,0 +1,44 @@
...
+from ..spiders.CustomSpider import CustomSpider
+
+class DeAnzaSpider(CustomSpider):
+    name = "deanza"
+
+    def start_requests(self):
+        start_url = "http://ecms.deanza.edu/deptoutlinespublic.html"
+
+        def get_departments(response):
+            for option in response.css("#Dept option"):
```
jodizzle: The first option in the department list looks like it has an empty value, so you should skip it, e.g.,
```python
for option in response.css("#Dept option")[1:]:
```

HarrisonGregg: Good catch.

osp_scraper/spiders/deanza.py
```diff
@@ -0,0 +1,45 @@
...
+                code = option.css("option::attr(value)").extract_first()
+                if code:
+                    name = option.css("option::text").extract_first()
```
jodizzle: Is this variable ever used?

jodizzle:
OK to merge (though you might want to clean up trailing whitespace and the like).

PR #35, opened 2017-03-20
Added functional GATech spider

HarrisonGregg:


osp_scraper/spiders/gatech.py
```diff
@@ -0,0 +1,48 @@
...
+                    meta={
+                        'source_url': response.url,
+                        'source_anchor': name,
+                        'depth': 1,
+                        'hops_from_seed': 1,
+                    },
+                    callback=self.parse_for_files,
+                )
+
+        for url in GATechSpider.start_urls:
```
jodizzle: We typically use `start_requests` for consistency, but for cases like this, I'm wondering if it would make more sense to use `parse` instead and be handed the responses from the `start_urls` list automatically.  You also wouldn't need a nested function in that case.

osp_scraper/spiders/gatech.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class GATechSpider(CustomSpider):
+    name = "gatech"
+
+    start_urls = [
```
jodizzle: How much of GATech does the list here cover?  Are there other parts that need separate scrapers?

HarrisonGregg: Not that I know of.  I think the rest of the site is handled by the standard scraper.

osp_scraper/spiders/gatech.py
HarrisonGregg: Good point.  I changed it to implement `parse`.

osp_scraper/spiders/gatech.py
```diff
@@ -0,0 +1,44 @@
...
+        "http://www.modlangs.gatech.edu/courses",
+    ]
+
+    def parse(self, response):
+        for option in response.css("select[name='iacTermcode'] option"):
+            code = option.css("option::attr(value)").extract_first()
+            name = option.css("option::text").extract_first()
+
+            yield scrapy.FormRequest(
+                response.url,
```
jodizzle: Minor: Trailing whitespace.

jodizzle:
OK to merge.

PR #37, opened 2017-03-20
Added functional ILCC spider

HarrisonGregg:


osp_scraper/spiders/ecollege.py
```diff
@@ -0,0 +1,43 @@
...
+    name = "ecollege"
+
+    def start_requests(self):
+        start_url = "https://secure.ecollege.com/iowacc/index.learn?action=Catalog"
+
+        def get_terms_codes(response):
+            for option in response.css("form input[type='Radio']"):
+                code = option.css("input::attr(value)").extract_first()
+
+                if code:
```
jodizzle: Does `not code` ever happen?

osp_scraper/spiders/ecollege.py
```diff
@@ -0,0 +1,43 @@
...
+                        formdata={
+                            "SID": code
+                        },
+                        callback=parse_results
+                    )
+
+        def parse_results(response):
+            for row in response.css('table+table tr'):
+                url = row.css('td:last-child a::attr(href)').extract_first()
+                if url:
```
jodizzle: Does `not url` ever happen?

osp_scraper/spiders/ecollege.py
```diff
@@ -0,0 +1,43 @@
...
+                    anchor = self.clean_whitespace(row.css('td::text').extract_first() or "")
```
jodizzle: I'm confused by this choice of anchor.  Could you explain?

jodizzle:
Also, as a side note, is this scraper for Iowa Lakes Community College, or for the Iowa Community College Online Consortium at large?  (Are they different institutions/organizations?)

osp_scraper/spiders/ecollege.py
HarrisonGregg: It's supposed to just be all of the text from the row, including the course number, course name, etc.

osp_scraper/spiders/ecollege.py
HarrisonGregg: Definitely on the first row, and maybe some rows on some pages are missing links.  I didn't check them all.  I like this pattern as it's an easy to understand and accomplishes the desired behavior in a variety of situations, such as a first row or option that's different from the rest, a table that only has links on some rows, or unexpected input.  I include it in case it is necessary or as just an extra check to make sure we're getting what we're expecting.

osp_scraper/spiders/ecollege.py
HarrisonGregg: It shouldn't.  Here, this serves as an extra check.

HarrisonGregg:
I think it's supposed to be for all schools in the Iowa Community College Online Consortium, but I'm not sure if it has all the courses from all of them.

osp_scraper/spiders/ecollege.py
jodizzle: Couldn't you just iterate over `response.css('table+table tr td:last-child a')` instead?  The end result seems like it would be equivalent to what you're doing here, and with one less iteration for each row that did not contain a syllabi link.  If the site were inconsistent, it could be possible that you start selecting links that aren't links to syllabi, but `if url` only checks for a `None`/empty string, so that wouldn't prevent the scraper from proceeding to either crash or download non-syllabi.  This pattern does check against the possibility of `a` tags with no `hrefs`, but that seems unlikely.  It's a similar situation with `if code` above.

I feel like that relates to a bigger issue: using something like `if code` doesn't solve the problem of checking input particularly robustly as we can still accept arbitrary strings, and if it does succeed in some way, it does so silently: we don't know what links were skipped and why.  I don't think it's that big a deal right now, but I think in the future (time permitting) it would be possible to develop a still simple but more rigorous pattern, probably by wrapping the request statements in `try`-`except` blocks and logging errors as they appear.

HarrisonGregg: Given that I'm using text from the whole row for the anchor, I'm not sure how I could use a selector like that one.  One thing to keep in mind about these scrapers is that the files are going to be treated the same as the files generated by the standard scraper, so it doesn't really matter if there are extra files.  This code currently tries to get a URL from where it expects it to be in the row and checks that it did get a URL.  Even if it didn't, it wouldn't be a big deal if it yielded requests with no URL. Scrapy would throw an error and move on to the next request.

osp_scraper/spiders/ecollege.py
HarrisonGregg: Decided to remove it.

HarrisonGregg:
@jodizzle Can you take another look now?

osp_scraper/spiders/ecollege.py
jodizzle: You're right that the selector you were using was appropriate.  I wasn't paying attention to the anchor you were getting.

You're also right about Scrapy just throwing an error and moving on.  I had done some lite testing a while back that made me think that Scrapy wouldn't handle bad URLs well and scrapers would crash because of that.  I double-checked and that's not true (it also wouldn't really make much sense if it did given what Scrapy is intended to accomplish).

That being said, I think your `if` check is appropriate here because sometimes there aren't `a` tags in the cells you're looking in, but as a general approach it seems problematic.  If something on the website subtly changed, for instance, I think I would rather Scrapy error and make some noise than have a scraper silently skip downloads.

jodizzle:
Looks good.

osp_scraper/spiders/ecollege.py
HarrisonGregg: My philosophy on the question of robustness to change is to try to support changes that might come from the data layer, i.e. someone entering data into the system without changing the underlying framework, but not to worry about any other changes.  If someone changes the system itself, then it's highly likely that the scraper will stop working, so there's no point trying to support a few very specific ways that they might change the system.  I feel like it's more likely that the href to the syllabus would end up looking different from what we expected than that an href to something else would end up in that cell, so doing any sort of check on the URL could be counter productive.

PR #38, opened 2017-03-20
Added functional ORU spider

HarrisonGregg:


osp_scraper/spiders/oru.py
```diff
@@ -0,0 +1,50 @@
...
+                        meta={
+                            'source_anchor': name
+                        },
+                        callback=get_departments
+                    )
+                    
+        def get_departments(response):
+            for option in response.css("select[name='dept'] option"):
+                code = option.css('option::attr(value)').extract_first()
+                if code:
```
jodizzle: Are these `if code` conditionals used to check against the possibility of the empty-value options at the beginning of the lists like "Select a Term"?  If so, I would just slice those entries off the list and save yourself some indentation, e.g.,
```python
for option in response.css("select[name='dept'] option")[1:]:
```

osp_scraper/spiders/oru.py
```diff
@@ -0,0 +1,50 @@
...
+                            'depth': 1,
+                            'hops_from_seed': 1
+                        },
+                        callback=self.parse_for_files
+                    )
+
+        yield scrapy.Request(get_syllabi_url, callback=get_terms)
+
+    def extract_links(self, response):
+        for a in response.css('.main-content form a'):
```
jodizzle: Minor: Instead of `a`, I would use `tag` as the variable name like you have in some other scrapers.

HarrisonGregg: Done.

osp_scraper/spiders/oru.py
HarrisonGregg: I used to slice them off that way, but I settled on this pattern to be as robust as possible.  It definitely isn't _necessary_, but will handle some unexpected cases more cleanly.

PR #40, opened 2017-03-20
Added functional Foothill scraper

HarrisonGregg:


osp_scraper/spiders/foothill.py
```diff
@@ -0,0 +1,81 @@
...
+
+------WebKitFormBoundary
+Content-Disposition: form-data; name="search"
+
+Search for Course Description
+------WebKitFormBoundary--
+"""
+
+OUTLINE_BODY = """
+------WebKitFormBoundary
```
jodizzle: Is there a reason why you define these big strings outside the class?

osp_scraper/spiders/foothill.py
```diff
@@ -0,0 +1,81 @@
...
+                        method="POST",
+                        body=SEARCH_BODY.format(value),
+                        headers={
+                            "Content-Type": "multipart/form-data; boundary=----WebKitFormBoundary"
+                        },
+                        callback=parse_results
+                    )
+                    
+        def parse_results(response):
+            for _input in response.css("form input[type='radio']"):
```
jodizzle: Why does this variable name have an underscore in front of it?  If it's because the name `input` is reserved in Python, I would just choose a different name entirely.

osp_scraper/spiders/foothill.py
```diff
@@ -0,0 +1,81 @@
...
+            for option in response.css("#Department option"):
+                value = option.css("option::attr(value)").extract_first()
+                if value:
+                    name = option.css("option::text").extract_first()
+                    yield scrapy.FormRequest(
+                        search_url, 
+                        method="POST",
+                        body=SEARCH_BODY.format(value),
+                        headers={
+                            "Content-Type": "multipart/form-data; boundary=----WebKitFormBoundary"
```
jodizzle: I think you can actually get away with using `scrapy.FormRequest.from_response`, or at least it worked when I tested it.  Something like this:
```python
scrapy.FormRequest.from_response(
    response,
    formdata={
        'act': 'yes',
        'Department': value,
        'search': 'Search for Course Description'
    },
    method='POST'
)
```
This way you don't have to format strings, and you also don't have to set the header manually.  It should read the necessary information from the HTML `form` elements.

HarrisonGregg:
@jodizzle Wow. Scrapy is amazing.  I didn't realize it could handle weird forms like that.  Looks a lot better now.

jodizzle:
Looks good.  OK to merge.

PR #41, opened 2017-03-20
Added functional SFASU scraper

HarrisonGregg:


osp_scraper/spiders/sfasu.py
```diff
@@ -0,0 +1,48 @@
...
+                            'hops_from_seed': 1
+                        },
+                        callback=self.parse_for_files
+                    )
+
+        yield scrapy.Request(start_url, callback=get_terms_codes)
+
+    def extract_links(self, response):
+        for row in response.css("#ctl00_ContentPlaceHolder1_termreport1_UpdatePanel1 tr"):
+            relative_url = row.css("a::attr(href)").extract_first()
```
jodizzle: I think I understand what you're going for, but I think it might be valuable in this case to be a bit more strict with the selectors.  Within each of the `tr` elements, there are up to three links `a` elements: a link to a syllabus, a link to a professor CV, and a javascript button that opens a popup window.  Only the syllabus link is of interest to us (we don't want CVs, right?), and sometimes, when it's not available, you sweep up the CV link instead, and possibly the javascript button if the CV weren't available.  However, looking at the site, the syllabus links are only ever going to be in the first column of each row.  So I feel like doing something like
```python
for row in response.css("#ctl00_ContentPlaceHolder1_termreport1_UpdatePanel1 tr td:first-child"):
```
would be better.

HarrisonGregg:
@jodizzle I've cleaned it up and added an explanation, but I don't think we're going to find a better solution to the long viewstate problem. Everything else about it look good?

osp_scraper/spiders/sfasu.py
```diff
@@ -0,0 +1,51 @@
...
+    start_urls = ["https://orion.sfasu.edu/courseinformation/"]
+
+    def parse(self, response):
+        for option in response.css("select option"):
+            code = option.css("::attr(value)").extract_first()
+            if code:
+                name = option.css("::text").extract_first()
+                yield scrapy.FormRequest.from_response(
+                    response,
+                    # start_url,
```
jodizzle: Is this comment here necessary?

jodizzle:
@HarrisonGregg The problem is pretty annoying, but I guess we'll make note of it and leave it for now.  Thanks for making a docstring about it.

osp_scraper/spiders/sfasu.py
HarrisonGregg: Nope.

PR #45, opened 2017-03-20
Added functional Galveston College scraper

HarrisonGregg:


osp_scraper/spiders/gc.py
```diff
@@ -0,0 +1,85 @@
...
+                        "subject": subject
+                    }
+                    yield scrapy.FormRequest(
+                        get_syllabi_url, 
+                        method="POST",
+                        formdata=formdata,
+                        meta=formdata,
+                        callback=get_courses
+                    )
+                    return
```
jodizzle: Is there a reason for this `return`?

osp_scraper/spiders/gc.py
```diff
@@ -0,0 +1,85 @@
...
+                        "course": response.meta["course"], 
+                        "section": section
+                    }
+                    yield scrapy.FormRequest(
+                        get_syllabi_url, 
+                        method="POST",
+                        formdata=formdata,
+                        meta={
+                            "depth": 1,
+                            "hops_from_seed": 1,
```
jodizzle: Should both `depth` and `hops_from_seed` be 1 here?

osp_scraper/spiders/gc.py
HarrisonGregg: Nope. Good catch.

osp_scraper/spiders/gc.py
HarrisonGregg: Do you have reason to believe they shouldn't be? I have literally no idea what they are actual used for.

jodizzle: I think you should probably set the `depth` and `hops_from_seed` to 1 in `get_subjects`, and increment each by 1 for each successive post request.

HarrisonGregg:
@jodizzle I was able to refactor it to simplify it even further because of how from_response handles things.  How does it look?

osp_scraper/spiders/gc.py
```diff
@@ -0,0 +1,61 @@
...
+                },
+                callback=self.parse_for_sections
+            )
+
+    def parse_for_sections(self, response):
+        term = response.css("#term::attr(value)").extract_first()
+        subject = response.css("#subject::attr(value)").extract_first()
+        course = response.css("#course::attr(value)").extract_first()
+        for option in response.css("#section option"):
+            section = option.css("option::attr(value)").extract_first()
```
jodizzle: Can't you just use `::attr(value)`

jodizzle:
This looks very clean now.

PR #46, opened 2017-03-24
Add Illinois State University scraper

jodizzle:
There were slight differences in the requests for `syllabi/Database/` and `syllabi/Archive/`, but they were annoying enough that I decided to handle the differences explicitly with two different methods.

### DIFF
--- a/osp_scraper/spiders/illinoisstate.py
+++ b/osp_scraper/spiders/illinoisstate.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+
+import itertools
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class IllinoisStateSpider(CustomSpider):
+    name = "illinoisstate"
+    allowed_domains = ["illinoisstate.edu"]
+
+    def start_requests(self):
+        database_url = "https://casit.illinoisstate.edu/syllabi/Database/QuerySelect"
+        archive_url = "https://casit.illinoisstate.edu/syllabi/Archive/QuerySelect"
+
+        def get_database_searches(response):
+            depts = response.css('#Department option::attr(value)').extract()
+            sems = response.css('#semester option::attr(value)').extract()
+            years = response.css('#year option::attr(value)').extract()
+            for dept, sem, year in itertools.product(depts, sems, years):
+                yield scrapy.FormRequest(
+                    'https://casit.illinoisstate.edu/syllabi/Database/QueryResults',
+                    formdata={
+                        'Department': dept,
+                        'semester': sem,
+                        'year': year,
+                    },
+                    method='POST',
+                    meta={
+                        'depth': 1,
+                        'hops_from_seed': 1,
+                        'source_url': response.url,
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/illinoisstate.py
```diff
@@ -0,0 +1,78 @@
...
+class IllinoisStateSpider(CustomSpider):
+    name = "illinoisstate"
+    allowed_domains = ["illinoisstate.edu"]
+
+    def start_requests(self):
+        database_url = "https://casit.illinoisstate.edu/syllabi/Database/QuerySelect"
+        archive_url = "https://casit.illinoisstate.edu/syllabi/Archive/QuerySelect"
+
+        def get_database_searches(response):
+            departments = response.css('form select#Department option::attr(value)')
```
HarrisonGregg: I know you have to balance between being concise and being explicit, but in cases like this I think you're being a little too verbose.  IDs are unique in HTML, so there's no reason to include `form select`.

osp_scraper/spiders/illinoisstate.py
```diff
@@ -0,0 +1,78 @@
...
+            semesters = response.css('form select#semester option::attr(value)')
+            years = response.css('form select#year option::attr(value)')
+            searches = itertools.product(departments.extract(),
```
HarrisonGregg: Is there a reason you `extract` here instead of when the variable is defined?

osp_scraper/spiders/illinoisstate.py
jodizzle: Yeah I guess that makes sense, especially since the ID names are very indicative of the contents (unlike many pages where the IDs are less coherent).

osp_scraper/spiders/illinoisstate.py
jodizzle: Mostly for line length concerns, and it seemed like it looked clean enough to do extracts inside the `itertools.product`.  Maybe with your suggestion above, though, I should move the `extract`s up.

jodizzle: I decided to keep the `extract`s inside the itertools call.  I think it still looks alright?

jodizzle:
@HarrisonGregg

osp_scraper/spiders/illinoisstate.py
HarrisonGregg: I gotta say, I don't think the `extract`s in the `product` call looks alright.  The pattern we use elsewhere is to just `extract` on the same line as the selector if you can, and it seems unnecessary to break that pattern just to save a few characters on the line length, especially when it leads to a broken line afterwards.  How about something like this:
```
depts = response.css('#Department option::attr(value)').extract()
sems = response.css('#semester option::attr(value)').extract()
years = response.css('#year option::attr(value)').extract()

for dept, sem, year in itertools.product(depts, sems, years):
```
It follows the usual pattern, it keeps the lines under 80 chars, and I think it looks a lot cleaner.

jodizzle: Agreed.  I think shortening the variable names makes a lot of sense.

PR #47, opened 2017-03-27
Adding script and scraper for downloading files listed in CSV

HarrisonGregg:
This adds a script that can be run to generate WARC files from a list of files in a CSV.  It creates `PageItems` for each of the Source URLs in the CSV with all of the Document URLs as `file_urls`, so that the Warc Pipelines will automatically download the files and generate WARCs.  This required a workaround of requesting a dummy url (http://example.com/), and yielding `PageItems` from the handler, as yielding `PageItems` directly from `start_urls` is not permitted.

### DIFF
--- a/osp_scraper/spiders/file_scraper.py
+++ b/osp_scraper/spiders/file_scraper.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+import csv
+from itertools import groupby
+from operator import itemgetter
+
+from ..spiders.CustomSpider import CustomSpider
+from ..items import PageItem
+
+class FileDownloader(CustomSpider):
+    name = "file_downloader"
+
+    # Workaround since yielding PageItems from start_requests isn't allowed
+    start_urls = ["http://example.com/"]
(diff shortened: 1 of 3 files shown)


osp_scraper/spiders/file_scraper.py
```diff
@@ -0,0 +1,59 @@
...
+from ..spiders.CustomSpider import CustomSpider
+from ..items import PageItem
+
+class FileDownloader(CustomSpider):
+    name = "file_downloader"
+
+    # Workaround since yielding PageItems from start_requests isn't allowed
+    start_urls = ["http://example.com/"]
+
+    def parse(self, response):
```
wearpants: This code could be much simplified using [itertools.groupby](https://docs.python.org/3/library/itertools.html#itertools.groupby) on `row`s.

I stubbed out this: https://gist.github.com/wearpants/fae802e1d2e2f9158b06e7121e877181

wearpants:
Cool, can you add a line in the README and i think we're good

PR #48, opened 2017-03-29
Minor changes for readability and robustness

HarrisonGregg:


osp_scraper/spiders/CustomSpider.py
```diff
@@ -19,6 +18,9 @@ def from_crawler(cls, crawler, *args, **kwargs):
         spider.filters = [Filter.compile('allow')]
         return spider
 
+    def parse(self, response):
+        return
```
jodizzle: Is this necessary?  If you're doing this just for clarity (related to the fact that we mostly use `start_requests`), then you might want to add a short docstring, something like "Implement in subclass for each site as needed."

osp_scraper/spiders/CustomSpider.py
```diff
@@ -35,14 +37,14 @@ def parse_for_files(self, response):
...
 
         for url, anchor in self.extract_links(response):
-            meta={
+            meta = {
                 'source_url': response.url,
                 'source_anchor': anchor,
-                'depth': response.meta['depth'] + 1,
-                'hops_from_seed': response.meta['hops_from_seed'] + 1,
+                'depth': response.meta.get('depth', 0) + 1,
+                'hops_from_seed': response.meta.get('hops_from_seed', 0) + 1,
```
jodizzle: I think I like the idea of this change.  One question: are we always going to want `depth`/`hops_from_seed` to be at least 1 (representing at least one link crawled from the start)?  I think the answer has been yes so far?

osp_scraper/spiders/CustomSpider.py
```diff
@@ -35,14 +37,14 @@ def parse_for_files(self, response):
...
             }
 
-            file_urls.append((url, meta))
+            file_urls.append((response.urljoin(url), meta))
```
jodizzle: I suppose this is fine, though I almost feel like I'd rather people just make sure to `urljoin` their own inputs.  It also makes what's going on in each subclass more explicit.

osp_scraper/spiders/CustomSpider.py
jodizzle: Also, it might be better to use something a bit more specific than `return`, like `raise NotImplementedError`.

HarrisonGregg:
@jodizzle How about now?

osp_scraper/spiders/CustomSpider.py
```diff
@@ -35,14 +41,14 @@ def parse_for_files(self, response):
...
         for url, anchor in self.extract_links(response):
-            meta={
+            meta = {
                 'source_url': response.url,
-                'source_anchor': anchor,
-                'depth': response.meta['depth'] + 1,
-                'hops_from_seed': response.meta['hops_from_seed'] + 1,
+                'source_anchor': self.clean_whitespace(anchor),
+                'depth': response.meta.get('depth', 1) + 1,
+                'hops_from_seed': response.meta.get('hops_from_seed', 1) + 1,
```
jodizzle: If we're gonna go with a strategy like this, I think you should also do something similar down in the `yield PageItem` section, except maybe with `response.meta.get('[source_url or hops_from_seed]', 0)`.  That would be could useful in case there's anything that needs to be parsed from the landing page, or just as an extra check.

HarrisonGregg: I think that would make sense, but in that case, shouldn't these be 0 too?

HarrisonGregg:
@jodizzle How about now?

osp_scraper/spiders/CustomSpider.py
jodizzle: Yep, you're right.

osp_scraper/spiders/CustomSpider.py
```diff
@@ -19,6 +16,19 @@ def from_crawler(cls, crawler, *args, **kwargs):
         spider.filters = [Filter.compile('allow')]
         return spider
 
+    def start_requests(self):
+        for request in super().start_requests():
+            request.meta['depth'] = 0
+            request.meta['hops_from_seed'] = 0
```
jodizzle: These changes aren't really necessary with the changes to `parse_for_files` below.  These `meta` entries are only really relevant for generating `PageItem`s in `parse_for_files`, and you already cover the issue of a dictionary entry not being set with `response.meta.get('depth', 0)` and `response.meta.get('hops_from_seed', 0)`.

I think it would probably be best to just leave a docstring `raise NotImplementedError`, like you do with `parse`.

osp_scraper/spiders/CustomSpider.py
```diff
@@ -31,18 +41,24 @@ def parse_for_files(self, response):
...
         file_urls = []
 
         for url, anchor in self.extract_links(response):
-            meta={
+            meta = {
                 'source_url': response.url,
-                'source_anchor': anchor,
-                'depth': response.meta['depth'] + 1,
-                'hops_from_seed': response.meta['hops_from_seed'] + 1,
+                'source_anchor': self.clean_whitespace(anchor),
```
jodizzle: Should you clean the anchor down where you `yield PageItem` as well?

osp_scraper/spiders/CustomSpider.py
```diff
@@ -31,18 +41,24 @@ def parse_for_files(self, response):
...
+                'depth': response.meta.get('depth', 0) + 1,
+                'hops_from_seed': response.meta.get('hops_from_seed', 0) + 1,
             }
 
-            file_urls.append((url, meta))
+            file_urls.append((response.urljoin(url), meta))
+
+        if response.meta.get('require_files') and not file_urls:
+            return
```
jodizzle: I like this addition and think it will be helpful.  Could you add a comment about its intended purpose?

osp_scraper/spiders/CustomSpider.py
HarrisonGregg: Maybe we should leave this one and remove the `get`s with defaults below, so that an error is thrown if you try to save `PageItem`s without `depth` or `hops_from_seed`.

osp_scraper/spiders/CustomSpider.py
HarrisonGregg: In addition to the explanation in the docstring?

HarrisonGregg:
@jodizzle I decided to change it around so that it will complain about missing meta parameters if you try to run a spider without them.

osp_scraper/spiders/CustomSpider.py
jodizzle: Well if this `start_requests` parent method was implemented like it is here, wouldn't it cover that case anyway for when the user uses `start_requests`?  And if you have `start_requests` work like this, shouldn't you have `parse` work like this as well?

It seems like some of what you're trying to do, like with the changes in the `extract_links` call in `parse_for_files`, is use default values to prevent certain errors from occurring.  The alternative to me seems to be to let the user be responsible for everything.  Which do you want?

jodizzle:
@HarrisonGregg I added a docstring describing what the new `start_requests` does and tweaked another docstring.  Feel free to take a look and then rebase down to however many commits you're comfortable with.

osp_scraper/spiders/CustomSpider.py
jodizzle: I suppose what you've done here is pretty convenient in certain situations while not hiding mistakes in others.  If a user is `callback`ing to `parse_for_files` directly out of `parse` or making recursive calls from within `parse`, then this is a helpful setting, but a user still has to set the `meta` content correctly when `callback`ing from other functions.

I added a docstring about the behavior.

PR #51, opened 2017-04-04
Add St. Cloud Technical Community College scraper

jodizzle:
https://webproc.mnscu.edu/registration/viewCourseOutlines.do?campusid=208

Two option form.  Pretty straightforward.

### DIFF
--- a/osp_scraper/spiders/sctcc.py
+++ b/osp_scraper/spiders/sctcc.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SCTCCSpider(CustomSpider):
+    name = "sctcc"
+
+    start_urls = ["https://webproc.mnscu.edu/registration/viewCourseOutlines.do?campusid=208"]
+
+    def parse(self, response):
+        sem_values   = response.css("#selectedYearTerm option::attr(value)")\
+                               .extract()
+        sem_anchors  = response.css("#selectedYearTerm option::text")\
+                               .extract()
+        subj_values  = response.css("#selectedSubject option::attr(value)")\
+                               .extract()
+        subj_anchors = response.css("#selectedSubject option::text")\
+                               .extract()
+
+        for sem_value, sem_anchor in zip(sem_values, sem_anchors):
+            for subj_value, subj_anchor in zip(subj_values, subj_anchors):
+                yield scrapy.FormRequest(
+                    "https://webproc.mnscu.edu/registration/viewCourseOutlines.do",
+                    formdata={
+                        'selectedYearTerm': sem_value,
+                        'selectedSubject': subj_value
+                    },
+                    method="POST",
+                    meta={
+                        'depth': 1,
+                        'hops_from_seed': 1,
+                        'source_url': response.url,
+                        'source_anchor': sem_anchor + ' ' + subj_anchor,
+                        'require_files': True
+                    },
+                    callback=self.parse_for_files
+                )
+
+    def extract_links(self, response):
+        results = response.css("#searchResults tr td:last-child a")
+        for result in results:
+            rel_url = result.css("a::attr(href)").extract_first()
+            url = response.urljoin(rel_url)
+            anchor = result.css("a::attr(title)").extract_first()
+            yield (url, anchor)


osp_scraper/spiders/sctcc.py
```diff
@@ -0,0 +1,47 @@
...
+
+        def get_searches(response):
+            sem_values = response.css('#selectedYearTerm option::attr(value)')
+            sem_anchors = response.css('#selectedYearTerm option::text')
+            subj_values = response.css('#selectedSubject option::attr(value)')
+            subj_anchors = response.css('#selectedSubject option::text')
+            semesters = zip(sem_values.extract(), sem_anchors.extract())
+            subjects = zip(subj_values.extract(), subj_anchors.extract())
+
+            for semester, subject in itertools.product(semesters, subjects):
```
jodizzle: So doing it your way, I think the code would look like the below, right?

```python
sem_values = response.css('#selectedYearTerm option::attr(value)').extract()
sem_anchors = response.css('#selectedYearTerm option::text').extract()
subj_values = response.css('#selectedSubject option::attr(value)').extract()
subj_anchors = response.css('#selectedSubject option::text').extract()

for sem_value, sem_anchor in zip(sem_values, sem_anchors):
    for subj_value, subj_anchor in zip(subj_value, subj_anchor):
        ...
```

So is the issue that currently I define variables `semesters` and `subjects` to hold the `zip` declarations?  Or that I iterate over the zip declarations with `semester` and `subject`, which I then have to index?  While what I have currently is probably a little fancy, I think it reads pretty cleanly, and I don't really feel like I'm contorting my code.

Note that another direction is I could use a statement like
```python
for (sem_value, sem_anchor), (subj_value, subj_anchor), subject in itertools.product(semesters, subjects):
```
But that gets in to line length concerns.

HarrisonGregg: I would definitely go with the first one you have, but breaking the first 4 lines at `extract` if you want.  My problem with the current code is that the individual lines aren't as self explanatory as they could be.  Obviously, individual lines of code won't always be self explanatory, but here they can be.  
For example, to understand what's at index 0 of `semester`, you have to look back to the for loop, figure out that `semester` is coming from `semesters`, then look back to the declaration of `semesters` to find out that index 0 is the semester value.  Written as you have it above, you can just see that it's the semester value.  Why burden the reader with thinking about how zip creates a generator of tuples when you can just stick them on the same line within the loop, and the reader can recognize the pattern of `for a, b in zip(as, bs)`, and understand immediately that you're just iterating through two lists in parallel.  
It also bothers me to pass around generators of selectors, when you could just pass around a list of list of strings by extracting it immediately.  Again, the reader has to think about how the variables you're passing to the zip call aren't just lists, they're generators that need to be extracted before use.  
Individually, these things aren't that big of a deal, but all together, they just serve to distract the reader from what's really going on by subtly breaking the expected patterns that you can understand without thinking about them.

jodizzle: To your point about the selectors: technically, the CSS selectors we use aren't generators in the Python sense.  A command like `response.css('#selectedYearTerm option::attr(value)')` produces an instance of the `SelectorList` class, which is like a regular Python list but with some bonuses, like a `.extract()` method.  Correspondingly, it doesn't feel that unnatural me to carry around these objects without `.extract()`ing them right away; in fact, we do it in a lot of our scrapers, and it seems like a pretty normal part of working with Scrapy (though my decisions of when to extract in this scraper are pretty arbitrary).  Unless you meant something different?

That being said though, I think I agree with your other points and your overall sentiment.

jodizzle:
@HarrisonGregg I changed the style around a lot.

One of the things I realized while testing is that a lot of the semester/subject searches give empty results with no links to any syllabi.  However, because I make a call directly to `self.parse_for_files`, I still download those empty search result pages.  Is that problematic?

osp_scraper/spiders/sctcc.py
HarrisonGregg: Oh interesting.  I don't know why I assumed it was a generator.  Good to know.

HarrisonGregg:
I haven't been worrying about yielding unnecessary PageItems given that the general scraper produces a ton of extraneous WARC files.  But, if we wanted to prevent it, we could distinguish somehow between whether or not `parse_for_files` should save a PageItem if no links are extracted.  We could do this with a separate function or a flag or something.

osp_scraper/spiders/sctcc.py
```diff
@@ -0,0 +1,46 @@
...
+        sem_anchors  = response.css("#selectedYearTerm option::text")\
+                               .extract()
+        subj_values  = response.css("#selectedSubject option::attr(value)")\
+                               .extract()
+        subj_anchors = response.css("#selectedSubject option::text")\
+                               .extract()
+
+        for sem_value, sem_anchor in zip(sem_values, sem_anchors):
+            for subj_value, subj_anchor in zip(subj_values, subj_anchors):
+                yield scrapy.FormRequest(
```
HarrisonGregg: Did `from_response` not work here?

jodizzle: I suppose I could use `FormRequest.from_response`, but I avoided it because the POST request parameters were simple enough.

jodizzle:
@HarrisonGregg Looks like the `require_files` flag from #48 helps a lot.

PR #52, opened 2017-04-05
Add Victoria College scraper

jodizzle:
https://www.victoriacollege.edu/courseinstructorinformation

This one is fairly complicated.  There may be a way to reduce the number of requests, but at least with the approach I took, it's five callbacks.  There may be a few kinks to work out still.

This should be ready for review, but I'm leaving the dirty commit history in for reference.

### DIFF
--- a/osp_scraper/spiders/victoriacollege.py
+++ b/osp_scraper/spiders/victoriacollege.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class VictoriaCollegeSpider(CustomSpider):
+    name = "victoriacollege"
+
+    start_urls = ["https://www.victoriacollege.edu/courseinstructorinformation"]
+
+    def parse(self, response):
+        semesters = response.css("a[target='_blank'][title~='Courses']")
+        for semester in semesters:
+            url = semester.css("::attr(href)").extract_first()
+            anchor = semester.css("::text").extract_first()
+            yield scrapy.Request(
+                url,
+                callback=self.parse_subjects
+            )
+
+    def parse_subjects(self, response):
+        subjects = response.css("select#subj_id option::attr(value)").extract()
+        term_in = response.css("input[name='term_in']::attr(value)").extract_first()
+        yield scrapy.FormRequest.from_response(
+            response,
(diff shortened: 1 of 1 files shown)


HarrisonGregg:
Looks like you could make a scraper for [LSCO](https://lscossbprod.lsco.edu:9200/rgdb/bwckschd.p_disp_dyn_sched) that's pretty similar to this one.

osp_scraper/spiders/victoriacollege.py
```diff
@@ -0,0 +1,119 @@
...
+                        'source_url': response.url,
+                        'source_anchor': anchor,
+                    },
+                    callback=get_subjects
+                )
+
+        def get_subjects(response):
+            subjects = response.css('select#subj_id option::attr(value)').extract()
+            term_in = response.css('input[name="term_in"]::attr(value)').extract_first()
+            yield scrapy.FormRequest(
```
jodizzle: I believe I had tried using `from_response` and ran into problems, but I can check again.

osp_scraper/spiders/victoriacollege.py
```diff
@@ -0,0 +1,119 @@
...
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class VictoriaCollegeSpider(CustomSpider):
+    name = 'victoriacollege'
+    allowed_domains = ['victoriacollege.edu']
+
+    def start_requests(self):
+        start_url = "https://www.victoriacollege.edu/courseinstructorinformation"
```
jodizzle: I don't seem to be getting this, but I'm worried about this happening in the future.  Maybe it has to do with the versions of the packages in `requirements.txt`?

jodizzle:
@HarrisonGregg I originally tried to follow your advice of getting ["syllabus urls directly from the search results page"](https://github.com/opensyllabus/osp-scraper/pull/52#pullrequestreview-31643129), but after attempting that a bit in https://github.com/opensyllabus/osp-scraper/pull/52/commits/8f8704df0aa1a1fa04ff394989292f67b0a69ef8, it seems like its hard to select both the syllabus link and its appropriate anchor (especially cause sometimes the syllabus link won't exist).  I could probably nail it if I spent more time working on it, but I think it might just be easier to stick to the old method, even if it is one more request.

PR #53, opened 2017-04-08
Added campusconcourse scraper for sites with pdf/doc files

HarrisonGregg:
At this point, this is just for [Northampton](https://northampton.campusconcourse.com/search), which has download buttons for doc/pdf files instead of links to HTML pages.

### DIFF
--- a/osp_scraper/spiders/campusconcourse_with_files.py
+++ b/osp_scraper/spiders/campusconcourse_with_files.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class CampusConcourseWithFilesSpider(CustomSpider):
+    name = "campusconcourse_with_files"
+
+    def start_requests(self):
+        for start_url in self.start_urls:
+            yield scrapy.FormRequest(
+                start_url,
+                formdata={
+                    'search_performed': "1"
+                },
+                method="GET",
+                callback=self.parse_for_pages
+            )
+
+    def parse_for_pages(self, response):
+        pages_text = response.css(".panel-footer label::text").extract_first()
+        num_pages = int(pages_text.split()[-1])
+
+        for page in range(num_pages):
+            yield scrapy.FormRequest(
+                response.url,
+                formdata={
+                    'search_performed': "1",
+                    'offset': str(page)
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/campusconcourse_with_files.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class CampusConcourseWithFilesSpider(CustomSpider):
+    name = 'campusconcourse_with_files'
+
+    def start_requests(self):
```
jodizzle: Do you mean to have both a `start_requests` and a `parse` in the same file?

osp_scraper/spiders/campusconcourse_with_files.py
```diff
@@ -0,0 +1,49 @@
...
+                method='GET',
+                formdata={
+                    'search_performed': '1'
+                }
+            )
+
+    def parse(self, response):
+        pages_text = response.css('.panel-footer label::text').extract_first()
+        num_pages = int(pages_text.split()[-1])
+        print(num_pages)
```
jodizzle: Test code?

osp_scraper/spiders/campusconcourse_with_files.py
```diff
@@ -0,0 +1,49 @@
...
+                method='GET',
+                formdata={
+                    'search_performed': '1',
+                    'offset': str(page)
+                },
+                meta={
+                    'source_url': response.url,
+                    'source_anchor': 'Search Page ' + str(page),
+                    'depth': 0,
+                    'hops_from_seed': 1,
```
jodizzle: Set `depth` and `hops_from_seed` to be the same.

HarrisonGregg:
@jodizzle I think that it does work, scrapy just prioritizes the requests strangely.  It seems like it doesn't follow the redirects immediately, so it queues up a huge number of redirected requests before making the requests and saving the WARC files.  You can see it save them faster, if you limit the number of pages or files downloaded per page, but otherwise it will save them all eventually.

I made the other changes you requested though.  How does it look now?

jodizzle:
@HarrisonGregg I see, I've got it working now.  Looks ready to merge.

PR #56, opened 2017-04-09
Added functional Clark State scraper

HarrisonGregg:


osp_scraper/spiders/clarkstate.py
```diff
@@ -0,0 +1,76 @@
...
+                formdata={
+                    'term': term,
+                    'archives': "true"
+                },
+                method="GET",
+                meta={
+                    'term': term
+                },
+                headers={
+                    "Accept": "application/json"
```
HarrisonGregg: Really? Did you get it to work without it? The request fails when i try it without this in the shell.

PR #57, opened 2017-04-10
Added functional Amherst scraper

HarrisonGregg:


osp_scraper/spiders/amherst.py
```diff
@@ -0,0 +1,58 @@
...
+                meta={
+                    'depth': 1,
+                    'hops_from_seed': 1,
+                },
+                callback=self.parse_for_courses
+            )
+
+    def parse_for_courses(self, response):
+        for a_tag in response.css(".sched-results tr td:nth-child(3) a"):
+            anchor = a_tag.css("::text").extract_first()
```
jodizzle: It might be nice to include in the anchor the course code, which is in `td:nth-child(2)`.

HarrisonGregg: I had that at first, but it kept putting weird symbols into the anchor because of links that are sometimes in that cell also, so I just went with the course name.

PR #59, opened 2017-04-14
Added functional Santa Rosa scraper

HarrisonGregg:


osp_scraper/spiders/santarosa.py
```diff
@@ -0,0 +1,43 @@
...
+                callback=self.parse_for_courses
+            )
+
+    def parse_for_courses(self, response):
+        rows = response.css("#tblCourseVersions tr")[1:]
+        for row in rows:
+            relative_url = row.css("a::attr(href)").extract_first()
+            url = response.urljoin(relative_url)
+            course_number = row.css("a::text").extract_first()
+            course_title = row.css("td:nth-child(2) ::text").extract_first()
```
jodizzle: Is there a reason there's a space between `td:nth-child(2)` and `::text`?  Is that necessary when using the `:nth-child` modifier?

HarrisonGregg: I had a feeling you'd ask about that. The space is there because the text isn't necessarily on the `td` itself, it can be within a child of the `td`.  With the space, it can select either.

osp_scraper/spiders/santarosa.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def parse(self, response):
+        departments = ["AG", "AGBUS", "AGMEC", "AGRI", "AJ", "ANAT", "ANHLT", "ANSCI", "ANTHRO", "AODS", "APE", "APED", "APGR", "APTECH", "ARCH", "ART", "ASL", "ASTRON", "ATHL", "AUTO", "BAD", "BBK", "BEHSC", "BGN", "BIO", "BMG", "BMK", "BOT", "BOTANY", "CEST", "CFS", "CHEM", "CHILD", "CHLD", "CHW", "CI", "CIS", "CLTX", "COMM", "CONS", "COUN", "CS", "CSKL", "CSKLS", "CUL", "DA", "DANCE", "DE", "DET", "DH", "DIET", "DNA", "DRD", "ECON", "EDUC", "ELEC", "EMC", "ENGL", "ENGR", "ENVS", "ENVST", "ENVT", "EQSCI", "ERTHS", "ESL", "FASH", "FDNT", "FIRE", "FLORS", "FREN", "GD", "GEOG", "GEOL", "GERM", "GIS", "GUID", "HIST", "HLC", "HLE", "HORT", "HOSP", "HR", "HUMAN", "IED", "INDE", "INTDIS", "ITAL", "JOUR", "KAQUA", "KCOMB", "KFIT", "KINDV", "KINES", "KTEAM", "LIR", "LPE", "MA", "MACH", "MATH", "MEDIA", "METRO", "MICRO", "MUS", "MUSC", "MUSCP", "NR", "NRA", "NRM", "NRV", "OA", "PE", "PHARM", "PHIL", "PHYED", "PHYS", "PHYSC", "PHYSIO", "PLS", "POLS", "PSYCH", "RADT", "RE", "RELS", "SE", "SOC", "SOCS", "SPAN", "SPCH", "SRT", "SURV", "SUSAG", "THAR", "VE", "VIT", "WELD", "WEOC", "WINE", "WRKEX", "WTR", "WWTR"]
```
HarrisonGregg: I definitely think this is a case where having it all on one line is preferable.  You don't actually need to see all of the departments and putting them each on a line would add 130 lines.  I added a comment.

PR #62, opened 2017-04-16
Scraper/howard

HarrisonGregg:


osp_scraper/spiders/howard.py
```diff
@@ -0,0 +1,66 @@
...
+                    'hops_from_seed': 1,
+                    'source_url': response.url,
+                    'source_anchor': name,
+                    'require_files': True
+                },
+                callback=self.parse_results
+            )
+
+    def parse_results(self, response):
+        for row in response.css("table tr"):
```
jodizzle: You can skip the first row, since it's the table header.

osp_scraper/spiders/howard.py
```diff
@@ -0,0 +1,66 @@
...
+                        'depth': 2,
+                        'hops_from_seed': 2,
+                        'source_url': response.url,
+                        'source_anchor': anchor
+                    },
+                    callback=self.parse_for_files
+                )
+
+        for item in self.parse_for_files(response):
+            yield item
```
jodizzle: I'm confused.  Why do you do a `self.parse_for_files` here but also callback to `self.parse_for_files` inside the request above?  Is this related to how some syllabi are HTML and some are PDFs?

HarrisonGregg: Yeah, exactly. Some are HTML and some are PDFs, so we need to create a PageItem for the search page, and for the HTML syllabi.

osp_scraper/spiders/howard.py
```diff
@@ -0,0 +1,66 @@
...
+    obvious way to identify them without performing the request.  As they are
+    unsuccessful status codes, they won't be saved.
+    """
+
+    name = "howard"
+
+    start_urls = ["https://www.howard.edu/syllabi/syllabi2/search.aspx"]
+
+    def parse(self, response):
+        options = response.css("#_ctl0_ContentPlaceHolder1_SCHOOL option")[1:]
```
HarrisonGregg: There aren't very many, but why would we want to exclude courses that aren't designated to a specific college?

osp_scraper/spiders/howard.py
HarrisonGregg: I think I'd prefer not to copy the entire list with a splice if it's not required.

osp_scraper/spiders/howard.py
```diff
@@ -0,0 +1,66 @@
...
+        for option in options:
+            name = option.css("::text").extract_first()
+            code = option.css("::attr(value)").extract_first()
```
HarrisonGregg: Good call.

osp_scraper/spiders/howard.py
jodizzle: Oops, I didn't actually mean to submit this comment.  I realized while working on the review that "No College Designated" had search results associated with it.

HarrisonGregg:
@jodizzle How about now?

osp_scraper/spiders/howard.py
jodizzle: If you could add a comment about this, that would be nice.

osp_scraper/spiders/howard.py
jodizzle: It would be nice if you could avoid the first row, but I guess it's not that important.

PR #63, opened 2017-04-16
Add College of New Caledonia scraper

jodizzle:
http://tools.cnc.bc.ca/courseoutlines/Default.aspx

Fairly straightforward.  As mentioned in the docstring, the scraper seems to take a very long time to download files, but seems to work as intended.

### DIFF
--- a/osp_scraper/spiders/cnc.py
+++ b/osp_scraper/spiders/cnc.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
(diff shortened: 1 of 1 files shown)


HarrisonGregg:
@wearpants @davidmcclure Do either of you have any idea why scrapy would wait to follow redirects until it had completed every other request generated by the scraper? The expected behavior would be for it to follow all redirects immediately, as they should have higher priority than other new requests.  This just means that the scraper will get all of the actual file URLs from 302 responses, and then download the files after it has done everything else.

jodizzle:
Sounds good.

PR #64, opened 2017-04-16
Add functional UT Dallas scraper

HarrisonGregg:


osp_scraper/spiders/utdallas.py
```diff
@@ -0,0 +1,74 @@
...
+            "https://coursebook.utdallas.edu/clips/clip-syllabus.zog",
+            method="POST",
+            formdata={
+                'id': response.meta['course_id'],
+                'div': "r-1childcontent",
+                'action': "syllabus"
+            },
+            meta={
+                'depth': 1,
+                'hops_from_seed': 1,
```
jodizzle: I feel like `depth` and `hops_from_seed` should be at least 2 here.  I suppose in some sense you're only visiting one page, but the actions in `parse` simulate the pressing of the search button, which should count as one hop/one depth.

osp_scraper/spiders/utdallas.py
```diff
@@ -0,0 +1,74 @@
...
+            .css(".selectlist-block:nth-child(3) select option::attr(value)")\
+            .extract()
+        for term, prefix in itertools.product(terms, prefixes):
+            url = url_format.format(term, prefix)
+            yield scrapy.Request(url, callback=self.parse_for_courses)
+
+    def parse_for_courses(self, response):
+        for row in response.css("tbody tr"):
+            course_id = row.css("a::attr(onclick)")\
+                .re_first(r"open_subrow\(.*?,.*?,.*?, '(.*?)'")
```
jodizzle: This is fairly descriptive on its face, but if you could add a comment about what you're matching here, that would be nice.

osp_scraper/spiders/utdallas.py
```diff
@@ -0,0 +1,74 @@
...
+            anchor = row.css("td:nth-child(3)::text").extract_first()
```
jodizzle: Not super necessary, but I think the course number in `td:nth-child(2)` is more indicative of or as indicative of the specific course being targeted.

osp_scraper/spiders/utdallas.py
```diff
@@ -0,0 +1,74 @@
...
+            method="POST",
+            formdata={
+                'id': response.meta['course_id'],
+                'div': "r-1childcontent",
+                'action': "syllabus"
+            },
+            meta={
+                'depth': 1,
+                'hops_from_seed': 1,
+                'source_url': response.url,
```
jodizzle: Should the `source_url` for all the downloaded files really be https://coursebook.utdallas.edu/clips/clip-coursebook-overview.zog?  In terms of browsing the site normally, I feel like that URL is mostly embedded content, and the `source_url` would be the search URL (or, depending on how you did it, the "homepage" for each course, e.g. https://coursebook.utdallas.edu/biol2111.0u1.16u).

osp_scraper/spiders/utdallas.py
HarrisonGregg: Good point. I'll set the anchor to both.

osp_scraper/spiders/utdallas.py
HarrisonGregg: Why? I can't think of anything I can say about it that isn't shown in the code itself.  I could say "Get the course ID from the 4th parameter of the open_subrow function called in the onclick attribute", but all of that is already in the code.

osp_scraper/spiders/utdallas.py
HarrisonGregg: Okay.

osp_scraper/spiders/utdallas.py
HarrisonGregg: I guess it's kind of a gray area in terms of what it _should_ be. If you click on a link that opens an `iframe` and that `iframe` has a link, the `iframe` url would be the source url, but I'll change it to the search url given that it provides more useful information.

osp_scraper/spiders/utdallas.py
jodizzle: Maybe it's just me, but I feel like regular expressions are nice and compact but usually opaque enough to warrant some sort of comment.  For example, in the [Marylhurst](https://github.com/opensyllabus/osp-scraper/pull/67/files#diff-b4229db1dcf42b57b4a82aae555b2ac6R29) scraper you describe what the form of the target is, which I find helpful.

HarrisonGregg: I feel the opposite.  Regular expressions clearly show you what you're expecting, whereas something like [this](https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/spiders/tru.py#L34) is completely opaque:
```
url = onclick.split("'")[1]
```
I've added a comment anyways, but it seems completely redundant.

HarrisonGregg:
@jodizzle

osp_scraper/spiders/utdallas.py
jodizzle: In your Marylhurst scraper, you write:
```python
# onclick is of the form:
# return showCrsDescription('<code>', '<term>');
code, term = a_tag.css("::attr(onclick)").re("'(.*?)'")
```
Which I think is quite nice because without it, I feel like I would have trouble evaluating exactly what the regular expression was trying to do, whether it was effective or why it was necessary to write it that way.  So I don't think you need to write
```python
# Get the course ID from the fourth argument to open_subrow.
```
but rather just describe what an `onclick` element or a course ID looks like.

Personally, for the example you picked out, I wouldn't say its opaque per se (`split`ing and getting an index are both common Python operations), but it does look pretty random and perhaps also deserving of a comment (though the fact that the result is being assigned to a variable called `url` makes it pretty clear that the result looks like a URL).  The amount of documentation we should be applying to these scrapers is something I've been thinking about, especially given the various ways a spider could break/miss an input or a site could change in the future.

If leaving a comment bothers you, you can leave it off.

HarrisonGregg:
@jodizzle I made the comment more like what you were suggesting and fixed the typo.

PR #71, opened 2017-04-17
Scraper/brevard

HarrisonGregg:


osp_scraper/spiders/brevard.py
```diff
@@ -0,0 +1,67 @@
...
+                callback=self.parse_for_courses
+            )
+        elif self.search_type == "College Credit":
+            self.search_type = "Continuing Education"
+            self.page = 1
+            # Reset the fingerprints set, so that the scraper can revisit pages
+            # with the same url now that the cookies have been changed
+            self.crawler.engine.slot.scheduler.df.fingerprints = set()
+            yield scrapy.Request(
+                "http://web12.easternflorida.edu/ecpr/choice.cfm?CE=1"
```
jodizzle: Out of curiosity, is this really how it works?  You don't have to `callback` to anything?

osp_scraper/spiders/brevard.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def parse_for_courses(self, response):
+        for item in self.parse_for_files(response):
+            yield item
+
+        results = response.css("table table")[0].css("tr")[1:]
+        next_url = response\
+            .css("form > table:last-child a:last-child::attr(href)")\
+            .extract_first()
+        if results and next_url:
```
HarrisonGregg: I don't think there is a next url when the first page isn't full, which came up in testing, but that shouldn't actually happen when doing an empty search.

osp_scraper/spiders/brevard.py
```diff
@@ -0,0 +1,67 @@
...
+    page = 1
+
+    def parse(self, response):
+        yield scrapy.FormRequest.from_response(
+            response,
+            meta={
+                'depth': 1,
+                'hops_from_seed': 1,
+                'source_url': response.url,
+                'source_anchor': "College Credit Search " + str(self.page),
```
HarrisonGregg: Good call.

osp_scraper/spiders/brevard.py
HarrisonGregg: Parse is the default callback, it just doesn't come up very often in our custom scrapers.

osp_scraper/spiders/brevard.py
```diff
@@ -0,0 +1,67 @@
...
+                "http://web12.easternflorida.edu/ecpr/choice.cfm?CE=1"
+            )
+
+    def extract_links(self, response):
+        for row in response.css("table table")[0].css("tr")[1:]:
+            url = row.css(".text10::attr(href)").extract_first()
+            if url:
+                document_name = row.css(".text10::text").extract_first()
+                course_title = row.css("td:nth-child(2)::text").extract_first()
+                anchor = document_name + " " + course_title
```
HarrisonGregg: Oh good catch. I'll see if I can keep using the same course title in the loop until I find a new one.

HarrisonGregg:
That would work if scrapy really did a depth first search, but it doesn't seem like it does in practice.

osp_scraper/spiders/brevard.py
jodizzle: Oh interesting, I never realized.  It looks a bit weird to not see a callback, but I guess it's fine.

jodizzle:
Ah, right.  You're right.  I have a feeling that there is a particular way for Scrapy to handle this type of problem, but your approach is the best I've seen so far.

HarrisonGregg:
@jodizzle Finally fixed the few things you mentioned.

HarrisonGregg:
@jodizzle It turned out the HTML was malformed, so it wasn't getting parsed correctly.  I decided to just simplify the anchors to the actual text of the link instead of including the course name, since figuring out how to parse malformed HTML seemed like more trouble than it's worth.

jodizzle:
That's pretty annoying.  I suppose in that case I would prefer it if you would use XPath selectors or something to get the course names properly, but you're right, the anchor isn't that critical.

I'll approve your changes.

PR #73, opened 2017-04-18
Add Lamar State College-Orange scraper

jodizzle:
https://lscossbprod.lsco.edu:9200/rgdb/bwckschd.p_disp_dyn_sched

Very similar to Victoria College in #52.  The main differences are `parse` and `extract_links`, but even those are pretty slight.

It would be nice to make a generalized scraper class for these types of sites since we've run into a few of them.  On the other hand, there seem to be enough small differences between each that making a general class may not be worth it.

### DIFF
--- a/osp_scraper/spiders/lsco.py
+++ b/osp_scraper/spiders/lsco.py
@@ -1,0 +1,92 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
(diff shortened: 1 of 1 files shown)


jodizzle:
Unfortunately, I think the Victoria College site is just different enough in annoying enough ways to warrant a separate scraper.  I could be wrong though.  

I'll leave the issue for now, but its something to consider in the future.

PR #75, opened 2017-04-21
Add San Diego Community College District scraper

jodizzle:
http://schedule.sdccd.edu/index.cfm

This scraper probably took more time to produce than was worth it (the syllabi/booklists are pretty sparse), but I decided to finish it anyway.  The contents and my commentary are pretty verbose, but I could slim it down if necessary.

### DIFF
--- a/osp_scraper/spiders/sdccd.py
+++ b/osp_scraper/spiders/sdccd.py
@@ -1,0 +1,162 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SDCCDSpider(CustomSpider):
+    name = "sdccd"
+
+    start_urls = ["http://schedule.sdccd.edu/index.cfm"]
+
+
+    def parse(self, response):
+        terms = response.css("select[name='trm'] option")[1:]
+        for term in terms:
+            value = term.css("::attr(value)").extract_first()
+            anchor = term.css("::text").extract_first()
+            yield scrapy.FormRequest.from_response(
+                response,
+                method="POST",
+                formdata={
+                    # 'resetfilters' seems to populate the different form fields
+                    # based on the value of 'trm'.
+                    'selectaction': "resetfilters",
+                    'trm': value
+                },
+                meta={
+                    'source_anchor': anchor,
+                    'trm': value
+                },
+                callback=self.parse_for_campus_and_subject
+            )
+
+    def parse_for_campus_and_subject(self, response):
+        campuses = response.css("select[name='campus'] option")[1:]
+        campus_values = campuses.css("::attr(value)").extract()
+        campus_anchors = campuses.css("::text").extract()
+        subjs = response.css("select[name='subj'] option")[1:]
+        subj_values = subjs.css("::attr(value)").extract()
+        subj_anchors = subjs.css("::text").extract()
+
+        for campus_value, campus_anchor in zip(campus_values, campus_anchors):
+            for subj_value, subj_anchor in zip(subj_values, subj_anchors):
+                anchor = ' '.join([response.meta['source_anchor'],
+                                   campus_anchor,
+                                   subj_anchor])
+                yield scrapy.FormRequest.from_response(
+                    response,
+                    method="POST",
+                    formname="ClassSelectForm",
+                    formdata={
+                        'selectaction': "listschedule",
+                        'trm': response.meta['trm'],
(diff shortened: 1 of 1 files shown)


HarrisonGregg:
I suppose there's no reason not to use `extract_files` like that.  Why are you searching with the term, college, subject, and course?  It seems like you can get results with the term and one other field, and searching with all of them returns 0 results sometimes.  If there is a good reason, you might want to add it to the docstring.

jodizzle:
The reason I use such a specific search is that I was having trouble trying to iterate over multi-page search results.  I could revisit it at some point, but filling in four search fields seems to (in my limited testing) eliminate the possibility of multi-page search results.

But you're right, I should add a docstring about it.

jodizzle:
Oops, just looked at the code again, and I remembered incorrectly:  I do have functionality in there for iterating over the multi-page results.  At the time, I think I thought it was just as good to do really specific searches.  Maybe I should change it?

osp_scraper/spiders/sdccd.py
```diff
@@ -0,0 +1,155 @@
...
+            yield (url, anchor)
+
+        books = response.css("span[title='Textbook']")
+        for book in books:
+            # The book link 'onclick' element looks like this:
+            # openbook('<crn>','<term code>', '<college code> ','<instructor>','<course initials>','<course number>','<course title>')
+            # The [1::2] is used to select only the parameters after splitting,
+            # skipping over the commas.
+            book_codes = book.css("::attr(onclick)")\
+                             .extract_first().split("'")[1::2]
```
HarrisonGregg: I know we disagree on this point, but I think this would be cleaner and easier to understand using regex:
```python
book_codes = book.css("::attr(onclick)").re("'(.*?)'")
```

osp_scraper/spiders/sdccd.py
```diff
@@ -0,0 +1,155 @@
...
+                )
+
+    def parse_for_courses(self, response):
+        # The courses options are populated based on the chosen subject, and so
+        # need to be selected in separate requests.
+        courses = response.css("select[name='crse'] option")[1:]
+        for course in courses:
+            value = course.css("::attr(value)").extract_first()
+            anchor = response.meta['source_anchor'] + " " + course.css("::text")\
+                                                                  .extract_first()
```
HarrisonGregg: Seems awkward to split like this here, especially given that both lines end up being over 80 characters anyways. Maybe something like this?
```python
            course_name = course.css("::text").extract_first()
            anchor = response.meta['source_anchor'] + " " + course_name
```

osp_scraper/spiders/sdccd.py
```diff
@@ -0,0 +1,155 @@
...
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SDCCDSpider(CustomSpider):
+    name = "sdccd"
+
+    start_urls = ["http://schedule.sdccd.edu/index.cfm"]
+
+    books_url = "http://schedule.sdccd.edu/book/book.cfm?crn={0}&term={1}&campus={2}&instructor={3}&dept={4}&course={5}&coursedescription={6}"
```
HarrisonGregg: Given that the contents of this URL are kind of important, you might consider breaking it up:
```python
    books_url = (
        "http://schedule.sdccd.edu/book/book.cfm"
        "?crn={0}"
        "&term={1}"
        "&campus={2}"
        "&instructor={3}"
        "&dept={4}"
        "&course={5}"
        "&coursedescription={6}"
    )
```
and personally I would put it in the function where it's used, since it's only used once.

osp_scraper/spiders/sdccd.py
jodizzle: I was actually having problems using regex, but I think it might've been because I was using `re_first` out of habit.  I think your approach looks cleaner.

osp_scraper/spiders/sdccd.py
jodizzle: Sure.

osp_scraper/spiders/sdccd.py
jodizzle: Ah, good point.  I don't know why I made it a class variable.

jodizzle:
I implemented your style changes.

As far as the docstring is concerned (from our comments earlier in this thread), I decided to rely on inline `NOTE`s for explanation.  This may not be a super good idea; I'm still thinking about how to appropriately document scrapers.

PR #76, opened 2017-04-23
Add The University of Texas at San Antonio scraper

jodizzle:
https://bluebook.utsa.edu/

This one took a little while to work out, but is actually pretty straightforward.  My implementation seems to work, but I admit that it's a bit messy.  Trying to deal with the long lines was pretty frustrating.

### DIFF
--- a/osp_scraper/spiders/utsa.py
+++ b/osp_scraper/spiders/utsa.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class UTSASpider(CustomSpider):
+    name = "utsa"
+
+    start_urls = ["https://bluebook.utsa.edu/"]
+
+    def parse(self, response):
+        subjs = response\
+            .css("#ctl00_MainContentSearchQuery_searchCriteriaEntry_CourseSubjectCombo_OptionList ")\
+            .css("li::text")\
+            .extract()
+
+        viewstate = response.css("#__VIEWSTATE::attr(value)").extract_first()
+        viewstategenerator = response.css("#__VIEWSTATEGENERATOR::attr(value)")\
+                                     .extract_first()
+        eventvalidation = response.css("#__EVENTVALIDATION::attr(value)")\
+                                  .extract_first()
+        for hidden_value, subj in enumerate(subjs, 1):
+            yield scrapy.FormRequest(
+                response.url,
+                method="POST",
+                # As far as I'm aware, using from_response always fails, and all
+                # the formdata below is necessary.
+                formdata={
+                    '__VIEWSTATE':
+                        viewstate,
+                    '__VIEWSTATEGENERATOR':
+                        viewstategenerator,
+                    '__EVENTVALIDATION':
+                        eventvalidation,
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$TextBox':
+                        subj,
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$HiddenField':
+                        str(hidden_value),
+                    '__EVENTTARGET':
+                        "ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchBtn",
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchTypeRBL':
+                        "SUBJECT",
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$AcademicDeptTitleCombo$HiddenField':
+                        "0",
+                    'ctl00$MainContent$mainContent1$TotalPages':
+                        "1",
+                    'ctl00$MainContent$mainContent1$TotalRows':
+                        "0",
+                    'ctl00$MainContent$mainContent1$CurrentPage':
+                        "1",
+                    'ctl00$FooterBtnCntrl$feedbackRating_RatingExtender_ClientState':
+                        "0",
+                    'hiddenInputToUpdateATBuffer_CommonToolkitScripts':
+                        "1"
+                },
+                meta={
+                    'depth': 1,
+                    'hops_from_seed': 1,
+                    'source_url': response.url,
+                    'source_anchor': subj + " Page 1",
+                    'require_files': True,
+                    'subj': subj
+                },
+                callback=self.parse_for_pages
+            )
+
+    def parse_for_pages(self, response):
+        for item in self.parse_for_files(response):
+            yield item
+
+        main_table = response\
+            .css("#ctl00_MainContent_mainContent1_AccordionRndPanel")
+        # The next page button is always present, but if on the final page, it
+        # is given the "disabled" attribute.
+        next_page_button_disabled = response\
+            .css("#ctl00_MainContent_mainContent1_PagerImgBtn_NextTOP")\
+            .css("::attr(disabled)")
+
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/utsa.py
```diff
@@ -0,0 +1,105 @@
...
+
+from ..spiders.CustomSpider import CustomSpider
+
+class UTSASpider(CustomSpider):
+    name = "utsa"
+
+    start_urls = ["https://bluebook.utsa.edu/"]
+
+    def parse(self, response):
+        subjs = response.css("#ctl00_MainContentSearchQuery_searchCriteriaEntry_CourseSubjectCombo_OptionList li::text")\
```
HarrisonGregg: You can put a line break before the first `.css` and put the `li::text` selector in another `.css` on a different line.

osp_scraper/spiders/utsa.py
```diff
@@ -0,0 +1,105 @@
...
+
+        for hidden_value, subj in enumerate(subjs, 1):
+            yield scrapy.FormRequest(
+                response.url,
+                method="POST",
+                # As far as I'm aware, using from_response always fails, and all
+                # the formdata below is necessary.
+                formdata={
+                    '__EVENTTARGET': "ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchBtn",
+                    '__VIEWSTATE': response.css("#__VIEWSTATE::attr(value)").extract_first(),
```
HarrisonGregg: Given that these are the same every line, you should probably extract them before the loop, which would also help with line length.

osp_scraper/spiders/utsa.py
```diff
@@ -0,0 +1,105 @@
...
+                    '__VIEWSTATEGENERATOR': response.css("#__VIEWSTATEGENERATOR::attr(value)").extract_first(),
+                    '__EVENTVALIDATION': response.css("#__EVENTVALIDATION::attr(value)").extract_first(),
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchTypeRBL': "SUBJECT",
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$TextBox': subj,
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$HiddenField': str(hidden_value),
```
HarrisonGregg: You could either shorten these lines with something like 
```python
base = "ctl00$MainContentSearchQuery$searchCriteriaEntry$"
```
or you could put a line break after the `:` when there's actually a dynamic value.

osp_scraper/spiders/utsa.py
```diff
@@ -0,0 +1,105 @@
...
+                        .extract()
+
+        for hidden_value, subj in enumerate(subjs, 1):
+            yield scrapy.FormRequest(
+                response.url,
+                method="POST",
+                # As far as I'm aware, using from_response always fails, and all
+                # the formdata below is necessary.
+                formdata={
+                    '__EVENTTARGET': "ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchBtn",
```
HarrisonGregg: I can understand leaving the formdata parameters in the same order as they are in the original request, but it might make more sense to put all of the ones with dynamic values at the top or the bottom, so they're easier to see.

osp_scraper/spiders/utsa.py
```diff
@@ -0,0 +1,105 @@
...
+
+    start_urls = ["https://bluebook.utsa.edu/"]
+
+    def parse(self, response):
+        subjs = response.css("#ctl00_MainContentSearchQuery_searchCriteriaEntry_CourseSubjectCombo_OptionList li::text")\
+                        .extract()
+
+        for hidden_value, subj in enumerate(subjs, 1):
+            yield scrapy.FormRequest(
+                response.url,
```
HarrisonGregg: Not sure it's worth the trouble, but I got `from_response` to work like this:
```python
            request = scrapy.FormRequest.from_response(
                response,
                method="POST",
                formdata={
                    '__EVENTTARGET': "ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchBtn",
                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$TextBox': 
                        subj,
                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$HiddenField': 
                        str(hidden_value),
                },
                meta={
                    'depth': 1,
                    'hops_from_seed': 1,
                    'source_url': response.url,
                    'source_anchor': subj + " Page 1",
                    'require_files': True,
                    'subj': subj
                },
                callback=self.parse_for_pages
            )
            yield request.replace(
                 body=request.body.replace(b"&ctl00%24HeaderBtnCntrl%24HomeBtn=home", b"")
            )
```

osp_scraper/spiders/utsa.py
jodizzle: Agreed.

osp_scraper/spiders/utsa.py
jodizzle: Agreed.  I'm also going to add line breaks and turn some of the values into variables to help with line length.

osp_scraper/spiders/utsa.py
jodizzle: Hmm.  I think I want to avoid using variables as part of the key values, at least when dictionaries are defined this way.  There may be smarter, python-tricky ways to go about creating `formdata` dictionaries that (appear to) need to be formatted like this.

For now, I'll just add line breaks.

osp_scraper/spiders/utsa.py
jodizzle: Sure.  The order of parameters can vary in requests anyway, even when using the site normally.

osp_scraper/spiders/utsa.py
jodizzle: This is good to know for this and other scrapers.  I think for right now, I'm going to skip this type of refactoring.  However, I think what you've done here is a good idea for a batch refactoring of scrapers in the future.

jodizzle:
@HarrisonGregg I think I addressed most of your style concerns.

osp_scraper/spiders/utsa.py
```diff
@@ -0,0 +1,116 @@
...
+                    '__VIEWSTATEGENERATOR': viewstategenerator,
+                    '__EVENTVALIDATION': eventvalidation,
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$TextBox':
+                        subj,
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$CourseSubjectCombo$HiddenField':
+                        str(hidden_value),
+                    '__EVENTTARGET':
+                        "ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchBtn",
+                    'ctl00$MainContentSearchQuery$searchCriteriaEntry$SearchTypeRBL':
+                    "SUBJECT",
```
HarrisonGregg: Nitpick, but this should either be indented or on the previous line.

HarrisonGregg:
@jodizzle I only pointed out a few of them, but if you're going to break some of the long lines like that, you should probably do it for all of them.

jodizzle:
@HarrisonGregg Fixed.

jodizzle:
@HarrisonGregg No, you're right, I was moving too fast.  I'll break the lines you mentioned and then merge.

(Stylistically I think it doesn't look so good to have to line break that often, but I'm not sure how else to cleanly get around the long selector strings.)

PR #78, opened 2017-04-26
Add tool for making a CSV diff

jodizzle:
(Having trouble thinking of a better title for this PR...)

This is a script that takes as input two CSV files - 'old' and 'new' - and produces a new CSV file that is a diff of the two with respect to certain columns.  Currently, it is set to produce a CSV suitable for use by `bin/edu_repo_crawler.py` by diffing the "Doc URLs" and "Mixed URLs" columns; every row that has new URLs in one of those columns is preserved.  Rows that contain an entry in the "Custom Scraper Name" column are always preserved as well.

This script still needs testing and cleaning up, but I'm opening it up for review.

### DIFF
--- a/bin/csv_diff.py
+++ b/bin/csv_diff.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+"""Create a CSV file that's a diff of two CSVs with respect to certain columns."""
+
+import collections
+import csv
+
+import click
+
+def extract_urls(s):
+    """return a list of clean URLs from a comma-separated string"""
+    urls = (u.strip() for u in s.split(','))
+    urls = (u for u in urls if u.startswith('http'))
+    return list(urls)
+
+@click.command()
+@click.argument('old_csv_file', type=click.Path(exists=True))
+@click.argument('new_csv_file', type=click.Path(exists=True))
+@click.argument('out_csv', type=click.File('w'))
+def main(old_csv_file, new_csv_file, out_csv):
+    with open(old_csv_file) as old_csv, open(new_csv_file) as new_csv:
+        old_reader = csv.DictReader(old_csv)
+        new_reader = csv.DictReader(new_csv)
+        # id_list will contain
+        #   1) The 'id's (first CSV column) of the rows that have been changed.
+        #   2) Differences between certain columns in the old and new CSVs.
+        id_list = collections.defaultdict(dict)
+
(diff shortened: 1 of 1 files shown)


jodizzle:
@HarrisonGregg My original approach had some pretty critical issues (it wasn't producing even close to the right diff), but now I think I've resolved them.

With this approach, rows from the new spreadsheet are iterated over, and if they are different in the right ways - i.e., contain different sets of URLs in some specified columns - the difference of the two sets (new minus old) is taken.  Then only the rows that were different, contain a "Custom Scraper Name" or (for safety) don't have a proper ID in the 'id' cell are written out to a new CSV.

I could probably refactor some of the code into functions, and might still, but for now I'm leaving it explicit.

HarrisonGregg:
@jodizzle

bin/csv_diff.py
```diff
@@ -17,76 +15,43 @@ def extract_urls(s):
...
+            old_row = old_rows.get(new_row['name'])
+
+            if old_row:
+                old_urls = set(
+                    extract_urls(old_row['Database URLs'])
+                    + extract_urls(old_row['Doc URLs'])
+                    + extract_urls(old_row['Mixed URLs'])
+                )
+
+                if "campusconcourse" not in new_row['Custom Scraper Name']:
```
jodizzle: You should probably add a comment above this (maybe a `NOTE`), especially because it's something we could need to adjust in the future if there are other custom scrapers that ever need to use database URLs.

bin/csv_diff.py
```diff
@@ -17,76 +15,43 @@ def extract_urls(s):
...
-        # TODO: Handle rows that are exclusively in the new_csv_file?
-        for new_id, new_item in new_by_id.items():
-            new_doc_url_set = set(extract_urls(new_item['Doc URLs']))
-            new_mixed_url_set = set(extract_urls(new_item['Mixed URLs']))
-            new_search_url_set = set(extract_urls(new_item['1st Search Result Page']))
-
-            scraper_name = new_item.get('Custom Scraper Name', '')
-            if scraper_name:
-                id_list[new_id]['Custom Scraper Name'] = scraper_name
+        old_rows = { row['name']: row for row in csv.DictReader(old_csv) }
```
jodizzle: Is there a reason you pad spaces after and before the enclosing curly braces?

PR #79, opened 2017-04-28
Add MappingSpider class

jodizzle:
MappingSpider is an extension of FilterSpider that creates filters for the spider based on the `start_urls` attribute.  By subclassing MappingSpider, we can create spiders in a manner similar to spiders
extended from CustomSpider.  However, instead of targeting specific URLs for conversion into PageItems, we instead pass URLs into FilterSpider's `parse` method, which treats those URLs as starting points for further crawling.

For certain sites, particularly sites where the path to syllabi is mostly GET requests behind a handful of form elements and/or POST requests, or sites where syllabi and other important information are distributed in odd ways, creating mappers can be more productive than creating custom spiders.

Lots of words for a tiny addition; the current implementation of this is actually simpler than I was originally thinking.  What I have here should work, though it is hard to properly test - I've added a mapper as an example.  I do have some concerns about how the filters are made, where they should be made and what they're made from, but those relate to structural questions about the repository as a whole that we may want to address later rather than sooner.  In any case, I think this code should be open for discussion.


osp_scraper/spiders/utexas.py
```diff
@@ -0,0 +1,34 @@
...
+import scrapy
+
+from ..spiders.MappingSpider import MappingSpider
+
+class UTexasMapper(MappingSpider):
+    name = "utexas"
+
+    start_urls = ["https://utdirect.utexas.edu/apps/student/coursedocs/nlogon/"]
+
+    def start_requests(self):
```
HarrisonGregg: Seems like this functionality would be replicated in most mappers, so why don't you put it in the MappingSpider?  You could implement `start_requests` to call a `self.map` method on all the urls.

jodizzle: Oh good idea.  I'll do that.

jodizzle:
@HarrisonGregg I added `start_requests` to MappingSpider.  Does the documentation I have look okay?

osp_scraper/spiders/MappingSpider.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def start_requests(self):
+        """
+        Override in each subclass when necessary.
+        """
+        for r in super().start_requests():
+            r.callback = self.start_map
+            yield r
+
+    def start_map(self, response):
```
wearpants: This comment is confusing. What does this method do?

wearpants: I mean, _I_ understand what it does, but the docstring doesn't explain it.

osp_scraper/spiders/utexas.py
```diff
@@ -0,0 +1,30 @@
...
+                    'course_type': ["In Residence", "Extension"],
+                    'search': "Search"
+                },
+                meta={
+                    'depth': 1,
+                    'hops_from_seed': 1,
+                    'source_url': response.url,
+                    'source_anchor': semester
+                },
+                callback=self.parse
```
wearpants: This is subtle, and needs to be clearly documented.... I wonder if we'd be better off renaming the base class `FilterSpider.parse` method to something else... thoughts?

wearpants:
This looks like the right approach, I'm glad it was so 

Let's move `make_params()` https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/tasks.py#L33 to a top-level `parameters.py`. We'll have more of these eventually & we're using them outside `tasks.py` now, so it's time for a module.

The `MappingSpider` name was confusing to me. It's more like a `PreambleSpider` or `PrefixSpider` or something. Thoughts?

osp_scraper/spiders/MappingSpider.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        spider.allowed_file_types = ALLOWED_FILE_TYPES
+        return spider
+
+    # TODO: Cleaner approach?
+    def __init__(self):
+        self.filters = make_params(self.start_urls)['filters']
```
HarrisonGregg: Shouldn't this be done in `from_crawler`?

osp_scraper/spiders/MappingSpider.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+from . import FilterSpider, ALLOWED_FILE_TYPES
+from ..tasks import make_params
+
+class MappingSpider(FilterSpider):
+    """An extension of FilterSpider that creates filters based on start_urls."""
```
HarrisonGregg: This docstring doesn't seem to describe the purpose of the class very well.

osp_scraper/spiders/utexas.py
HarrisonGregg: I think of `FilterSpider` as implementing the default crawling behavior, so having that be `parse` makes sense, then a subclass like this one overrides the initial behavior before dropping into the default behavior by using `parse` as a callback.

jodizzle: I agree with @HarrisonGregg.  That's how I envisioned it as well, and in that sense I think using `parse` as a final callback for mappers is a good approach.

Do we think that documentation in the `MappingSpider` and `docs/` would be able to make this approach clear enough?

osp_scraper/spiders/MappingSpider.py
jodizzle: I had originally approached it this way because I thought that I needed access to `self`, which is not available in `from_crawler` because its a class method.  But now I realize that's silly, and I can just access `start_urls` with `cls`.

jodizzle:
Sorry about the lite documentation.  When I was writing this, I felt the behavior was fairly straightforward, but I really should at least approach docstrings like was done in `CustomSpider`.

@wearpants I was thinking that working on this class was going to expose the need for some changes to the structure of the repository.  Should I open a ticket/pull request to discuss this issue separately?  I have some miscellaneous thoughts.

In terms of the name, I'm not super attached to `MappingSpider`.  I just chose it because it's the term we were using to discuss the idea (though I admit that even along those lines `MapperSpider` or `MapSpider` would be better).  One of the nice things about including the word 'map' is that it allows for sensible naming of the subclasses in a way that differentiates them from children of `CustomSpider`, e.g., `class UTexasMapper(MappingSpider)`, whereas I don't feel like you could have quite the same thing with `PreambleSpider` or `PrefixSpider`.  On the other hand, though, I agree that something like `PrefixSpider` is more consistent with the language used in the project.

jodizzle:
I elaborate in the docstrings.  Do they look alright now?

Also, quick thing I realized: I currently set `ROBOTSTXT_OBEY` to `False` (copy-paste from `CustomSpider`).  Is this problematic, given the broad crawls that each mapper may end up doing?  Or should we keep the setting and maybe use more restrictive filters?

osp_scraper/spiders/MappingSpider.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    custom_settings = {
+        'ROBOTSTXT_OBEY': False
+    }
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        spider.allowed_file_types = ALLOWED_FILE_TYPES
+        spider.filters = make_params(cls.start_urls)['filters']
```
HarrisonGregg: Does this work like this? I thought it had to be something like `kwargs["start_ulrs"]` instead of `cls.start_urls`.

jodizzle: Seems to.  As far as I know, `cls.start_urls` is accessing the class variable `start_urls`, similar to how we might do `self.start_urls` in instance methods to access the instance variable `start_urls` (and class and instance are the same for our purposes).

Is using `kwargs` preferable for some reason?

HarrisonGregg: Oh right. Never mind. I was thinking you might want it to support having the start_urls provided from the spreadsheet, but we just removed that.

HarrisonGregg:
I suppose it would be better not to ignore `robots.txt` by default.  There could be some sites that have a `robots.txt` that limits the crawling, but doesn't stop it altogether, that we want to respect.

osp_scraper/spiders/MappingSpider.py
jodizzle: Hmm, that's a good question.  I guess there's no particular reason that a situation like `campusconcourse` couldn't come up in the mappers.  In fact, it may have already come up with a couple of Blackboard sites.

Are you imagining something like this?
```python
spider.filters = make_params(cls.start_urls)['filters']
spider.filters.append(make_params(kwargs['database_urls'])['filters'])
```

HarrisonGregg: You should probably do `kwargs.get('database_urls', [])`, but otherwise yeah.

jodizzle:
I removed the `ROBOTSTXT_OBEY` setting.

osp_scraper/spiders/utexas.py
wearpants: yeah, ok now

wearpants:
> Let's move make_params() https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/tasks.py#L33 to a top-level parameters.py. We'll have more of these eventually & we're using them outside tasks.py now, so it's time for a module.

This still needs to happen.

wearpants:
@jodizzle @HarrisonGregg see the latest commit - I hoisted the file types to a class attribute, which makes `FilterSpider` easier to subclass. I think we can just extend `start_requests` in individual spiders, and so don't need a separate `MappingSpider` class.

I haven't had a chance to really test this, but I hope the intent makes sense.

jodizzle:
@wearpants The bigger issue with subclassing `FilterSpider` directly is that currently `FilterSpider` doesn't actually create any filters itself.  This is part of the issue that `MappingSpider` is trying to solve.

However, I think I agree with you that there isn't that much of a need for a separate `MappingSpider` class.  Instead, we should just focus on refactoring the structure of `FilterSpider` and `make_params`.  Currently, adding something like
```python
@classmethod
def from_crawler(cls, crawler, *args, **kwargs):
    spider = super().from_crawler(crawler, *args, **kwargs)
    spider.filters = make_params(getattr(cls, 'start_urs', []) + kwargs.get('start_urls', []))['filters']
    return spider
```
to `FilterSpider` would probably be sufficient, but I think the nature of `make_params` should be changed as well.  Rather than have a `parameters.py`, change `make_params` to just `make_filters` and move it into `filterware.py`, then edit any dependent pieces like `bin/edu_repo_crawler.py` accordingly.  Having a method like the current `make_params` that returns `start_urls` and `allowed_file_types` doesn't seem too helpful anyway.  Does that sound reasonable?

wearpants:
This sounds like the right direction, and would let us roll up customspider
as a filter spider subclass as well... Ill make some sketches in that
direction

jodizzle:
@wearpants I pushed two commits that summarize what I was thinking.

One of the annoying parts of implementing `start_requests` in individual scrapers/mappers is that `depth` and `hops_from_seed` need to be set to `0`, which is currently automated by `start_requests` in the superclasses.  So currently, either explicit setting of `depth` and `hops_from_seed` or something similar to this will be required in each mapper:
```python
    def start_requests(self):
        for r in super().start_requests():
            r.callback = self.start_map
            yield r
```
This could probably be addressed by adding a method to `FilterSpider` or by some other structural changes.

I don't really see the value in making `CustomSpider` a subclass of `FilterSpider` like you suggest, since `CustomSpider` instances do no filtering and several of the `CustomSpider` methods they make use of seem pretty distinct.  What would `CustomSpider` inherit from `FilterSpider`?

HarrisonGregg:
I can see both sides for whether `CustomSpider` should subclass `FilterSpider`.  If it did, that would allow us to not distinguish between "mappers" and full custom spiders.  Mappers would just be custom spiders that used `parse` from `FilterSpider` as a callback.  The potential problems I see are in regards to filters.  If we left custom scrapers set the filters to allow all, then a "mapper" would need to change the filters somehow so the scraper didn't get out of control.  If we don't leave custom scrapers as allow all, then we might need to change existing ones somehow so they don't break.

osp_scraper/spiders/__init__.py
```diff
@@ -62,6 +62,14 @@ class FilterSpider(BaseSpider):
 
     name = "osp_scraper_spider"
 
+    allowed_file_types = ALLOWED_FILE_TYPES
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        spider.filters = make_filters(getattr(cls, 'start_urls', []))
```
wearpants: Why the `getattr`? If `start_urls` isn't set by this point, it's an error.

osp_scraper/filterware.py
```diff
@@ -1,16 +1,82 @@
...
+    """Generate filters for a spider from a list of URLs.
+
+    * Allow paths with matching prefix to infinite depth
+    * Allow same hostname to max depth of 2
+    * Allow other domains to max depth of 1
+
+    Args:
+        seed_urls (list)
+    """
+    filters = []
```
wearpants: Wonder if we should make an empty `seed_urls` a `ValueError` exception?

osp_scraper/spiders/__init__.py
wearpants: Actually, shouldn't this be `spider.start_urls`, not `cls.start_urls`?

wearpants:
@HarrisonGregg , that's pretty much what I'm thinking re: "mappers" using `parse` from `FilterSpider`.

Maybe we add a method to `FilterSpider` that generates the filters, so that it could be extended/overridden in subclasses. Something like:

```
    @classmethod
    def from_crawler(cls, crawler, *args, **kwargs):
        spider = super().from_crawler(crawler, *args, **kwargs)
        spider.filters = cls.make_filters(spider.start_urls)
        return spider

    @classmethod
    def make_filters(cls, start_urls):
        ...
        return filters
```

Thoughts?

osp_scraper/filterware.py
HarrisonGregg: I think that would only make sense if we modified `edu_repo_crawler` so that it didn't call this if there were no urls.  We could have an institution with only a custom scraper and no other URLs, and that shouldn't raise an exception.

bin/edu_repo_crawler.py
```diff
@@ -47,8 +48,11 @@ def main(csv_file, local, institution):
 
                 if urls:
                     log.info("Found %d URLs for %s", len(urls), row['name'])
-                    params = make_params(urls)
-                    log.debug("Parameters: %r", params)
+                    params = {
+                        'start_urls': urls,
+                        'filters': make_filters(urls)
```
HarrisonGregg: Does `filters` need to be specified here anymore if it's set in `FilterSpider`?

osp_scraper/spiders/__init__.py
HarrisonGregg: I think it should actually be both, if we want the method to work for the standard scraper and mappers.  So something like:
```python
        spider.filters = make_filters(getattr(cls, 'start_urls', []) + kwargs.get('start_urls', []))
```

HarrisonGregg:
Alternatively, we could just put an `allow_all` flag in `FilterSpider`.  `FilterSpider` would set it to `False`, `CustomSpider` would set it to `True`, and mappers subclassing `CustomSpider` would set it to `False` again.

osp_scraper/spiders/__init__.py
```diff
@@ -165,7 +173,6 @@ def url_to_prefix_params(url):
 
     return {
```
HarrisonGregg: Is this function actually used anywhere?

bin/edu_repo_crawler.py
wearpants: No, it wouldn't.

osp_scraper/spiders/__init__.py
wearpants:
```
in/csv_prefix_crawler.py
8:from osp_scraper.spiders import url_to_prefix_params
13:            osp_scraper.tasks.crawl.delay(**url_to_prefix_params(row[0]))
```

wearpants:
Another option I considered was modify the pipelines to treat the *absence* of a `.filter` attribute on a spider as an "allow all" flag or always-true-filter. This would also let the pipeline play nice with other 3rd-party spiders. Needs more investigation though.

osp_scraper/spiders/__init__.py
HarrisonGregg: The changes in this PR make it so that specifying `allowed_file_types` and `filters` isn't necessary, so this function wouldn't really do anything there.  Is there a reason that `csv_prefix_crawler` uses different filters?  If so,  `spiders/__init__.py` would need to be changed, but if not, this function can be removed.

jodizzle: As far as I can tell, `url_to_prefix_params` makes filters that allow prefix crawling and deny everything else, which would make sense given the name `csv_prefix_crawler`.  Do we even use that script for anything anymore?

osp_scraper/spiders/__init__.py
jodizzle: Yep, this was my oversight.  The `edu_repo_crawler` script, for instance, won't have proper filters on the crawls it invokes without the `kwargs` addition.

osp_scraper/spiders/__init__.py
wearpants: i don't think so

jodizzle:
I pushed some more commits that I think address the miscellaneous suggestions.

This PR is pretty old and broad at this point, so I suggest that we freshen things up and keep the history clean by breaking it into several new PRs:

- One PR to address the moving of `make_params` -> `make_filters` and the related changes that streamline the filter creation process and remove unnecessary bits like `url_to_prefix_params`.  (This could be two PRs, depending on how it was done).  These changes should be mostly accomplished by the commits on this branch, and so we could just transplant that work to a new branch.
- One PR to change the class hierarchy of the spider classes in order to make room for the mapping spider idea.  This would still need substantial decision making.

Regarding the second bullet, the main (but not definitive) idea that seems to have surfaced (starting at @HarrisonGregg's [comment](https://github.com/opensyllabus/osp-scraper/pull/79#issuecomment-302301045)) is that we should have `CustomSpider` inherit from `FilterSpider`, and have mappers just be subclasses of `CustomSpider` that callback to the `FilterSpider` `parse` method and crawl according to a set of filters.  How to best accomplish this still needs some discussion; a couple approaches have been suggested in the comments.

Thoughts on this plan?

wearpants:
sounds good

PR #80, opened 2017-05-01
Tweaks for EDU Repo crawler

jodizzle:
General cleanup and improvements for `edu_repo_crawler.py`.  The two biggest things are the first two commits:

1) Since `start_urls` is defined in each custom scraper class, we shouldn't be trying to set the `start_urls` based on a cell spreadsheet.
2) Since the '1st Search Result Page' can contain URLs as well, include that as part of the URL collection.

Also, as I `NOTE` in an inline comment, should 'Database URLs' really be part of the big crawl?  'Database URLs' often contain URLs that will have very permissive prefixes (e.g., `https://www.kingsu.ca/academics/bachelor-programs`), and therefore will be likely to seed very long and broad crawls.

### DIFF
--- a/bin/edu_repo_crawler.py
+++ b/bin/edu_repo_crawler.py
@@ -35,10 +35,7 @@ def main(csv_file, local, institution):
             if not institution or institution == row['id']:
                 # run custom scraper, but only if not running locally
                 if not local and 'Custom Scraper Name' in row:
-                    params = {
-                        "start_urls": extract_urls(row['Database URLs'])
-                    }
-                    crawl_func(row['Custom Scraper Name'], **params)
+                    crawl_func(row['Custom Scraper Name'])
 
                 # find comma-separated URLs in these columns
                 urls = extract_urls(row['Doc URLs'])
--- a/bin/edu_repo_crawler.py
+++ b/bin/edu_repo_crawler.py
@@ -1,12 +1,7 @@
 #!/usr/bin/env python3
 
-import sys
-import os.path
 import csv
 import logging
-import itertools
-import urllib.parse
-import re
 
 import click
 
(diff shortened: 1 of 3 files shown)


bin/edu_repo_crawler.py
```diff
@@ -33,24 +26,23 @@ def main(csv_file, local, institution):
...
-                        "start_urls": extract_urls(row['Database URLs'])
-                    }
-                    crawl_func(row['Custom Scraper Name'], **params)
+                    crawl_func(row['Custom Scraper Name'])
 
-                # find comma-separated URLs in these columns
+                # Find comma-separated URLs in these columns.
                 urls = extract_urls(row['Doc URLs'])
                 urls.extend(extract_urls(row['Mixed URLs']))
+                # NOTE: Should database URLs be included?
```
HarrisonGregg: Yes, the database URLs should definitely be included.  I don't think we'll ever solve the problem of having some prefixes that are too broad.  It doesn't seem to have been too much of a problem so far.

jodizzle:
@HarrisonGregg I added the exporting of a `database_urls` parameter to help with scrapers like campusconcourse that need to get URLs from the spreadsheet.

bin/edu_repo_crawler.py
```diff
@@ -33,24 +26,28 @@ def main(csv_file, local, institution):
...
                     }
                     crawl_func(row['Custom Scraper Name'], **params)
+                    break
 
-                # find comma-separated URLs in these columns
+                # Find comma-separated URLs in these columns.
                 urls = extract_urls(row['Doc URLs'])
                 urls.extend(extract_urls(row['Mixed URLs']))
                 urls.extend(extract_urls(row['Database URLs']))
+                urls.extend(extract_urls(row['1st Search Result Page']))
```
HarrisonGregg: This line should be removed.

jodizzle:
@HarrisonGregg My bad.

PR #82, opened 2017-05-04
Add University of Georgia scraper

jodizzle:
https://syllabus.uga.edu/Browse.aspx
As mentioned in the docstring, there seem to be a number of 500 errors, but I'm not quite sure if that's my fault or not.

### DIFF
--- a/osp_scraper/spiders/uga.py
+++ b/osp_scraper/spiders/uga.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class UGASpider(CustomSpider):
+    name = "uga"
+
+    start_urls = ["https://syllabus.uga.edu/Browse.aspx"]
+
+    def parse(self, response):
+        yield scrapy.FormRequest.from_response(
+            response,
+            method="POST",
+            formdata={
+                # 'DEP1' corresponds to the "View History of Syllabus Files by
+                # Department" option.  This seems to give the most yield with
+                # the lowest amount of form-work.
+                'RadioButtonList1': "DEP1",
+                'Button1': "Submit"
+            },
+            callback=self.parse_for_departments
+        )
+
+    def parse_for_departments(self, response):
+        departments = response.css("#ddlDepartment option")[1:]
+        for dept in departments:
+            value = dept.css("::attr(value)").extract_first()
+            anchor = dept.css("::text").extract_first()
+            yield scrapy.FormRequest.from_response(
+                response,
+                method="POST",
+                formdata={
+                    'ddlDepartment': value
+                },
+                meta={
+                    'depth': 2,
+                    'hops_from_seed': 2,
+                    'source_url': response.url,
+                    'source_anchor': anchor
+                },
+                dont_click=True,
+                callback=self.parse_for_courses
+            )
+
+    def parse_for_courses(self, response):
+        # Syllabi are stored at the same URL behind javascript links that feed
+        # into ASP.NET POST requests.  So, rather than implement extract_links,
+        # we yield the response of each POST request into parse_for_files.
+        rows = response.css("#gridFileList tr")[1:]
+        for row in rows:
+            link = row.css("td:nth-child(6) a")
+            if link:
+                # Links look like this:
+                # javascript:__doPostBack('<eventtarget>','<eventargument>')
+                eventtarget, eventargument = link.css("::attr(href)")\
+                                                 .re(r"'(.*?)'")
+                anchor = " ".join([
+                    response.meta['source_anchor'],
+                    *row.css("td::text").extract()[:3],
+                    link.css("::text").extract_first()
+                ])
+                yield scrapy.FormRequest.from_response(
+                    response,
+                    method="POST",
+                    formdata={
+                        '__EVENTTARGET': eventtarget,
+                        '__EVENTARGUMENT': eventargument,
+                    },
+                    meta={
+                        'depth': 3,
+                        'hops_from_seed': 3,
+                        'source_url': response.url,
+                        'source_anchor': anchor
+                    },
+                    dont_click=True,
+                    callback=self.parse_for_files
+                )


osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,88 @@
...
+                anchor += " ".join(row.css("td::text").extract()[:3])
+                anchor += " " + link.css("::text").extract_first()
+                yield scrapy.FormRequest.from_response(
+                    response,
+                    method="POST",
+                    formdata={
+                        '__EVENTTARGET': eventtarget,
+                        '__EVENTARGUMENT': eventargument,
+                        'RadioButtonList1': "DEP1",
+                        'ddlDepartment': "AFS"
```
HarrisonGregg: I think removing this line should fix your 500 errors.

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,88 @@
...
+    def parse_for_departments(self, response):
+        departments = response.css("#ddlDepartment option")[1:]
+        for dept in departments:
+            value = dept.css("::attr(value)").extract_first()
+            anchor = dept.css("::text").extract_first()
+            yield scrapy.FormRequest.from_response(
+                response,
+                method="POST",
+                formdata={
+                    'RadioButtonList1': "DEP1",
```
HarrisonGregg: It seems like `from_response` grabs this automatically, so you don't need to include it here.

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,88 @@
...
+                    'source_url': response.url,
+                    'source_anchor': anchor
+                },
+                dont_click=True,
+                callback=self.parse_for_courses
+            )
+
+    def parse_for_courses(self, response):
+        # Grab the search result page.
+        for item in self.parse_for_files(response):
```
HarrisonGregg: There's no reason to grab the search result page if you're not implementing `extract_links`.

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,88 @@
...
+                response,
+                method="POST",
+                formdata={
+                    'RadioButtonList1': "DEP1",
+                    'ddlDepartment': value
+                },
+                meta={
+                    'depth': 2,
+                    'hops_from_seed': 2,
+                    'source_url': response.url,
```
HarrisonGregg: It doesn't look like you use this anywhere in the callback.

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,88 @@
...
+                anchor = response.meta['source_anchor']
+                anchor += " ".join(row.css("td::text").extract()[:3])
+                anchor += " " + link.css("::text").extract_first()
+                yield scrapy.FormRequest.from_response(
+                    response,
+                    method="POST",
+                    formdata={
+                        '__EVENTTARGET': eventtarget,
+                        '__EVENTARGUMENT': eventargument,
+                        'RadioButtonList1': "DEP1",
```
HarrisonGregg: I think you can remove this too.

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,88 @@
...
+                method="POST",
+                formdata={
+                    'RadioButtonList1': "DEP1",
+                    'ddlDepartment': value
+                },
+                meta={
+                    'depth': 2,
+                    'hops_from_seed': 2,
+                    'source_url': response.url,
+                    'source_anchor': anchor
```
HarrisonGregg: How about you call this something like `'department_name'` for clarity.

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,88 @@
...
+        for row in rows:
+            link = row.css("td:nth-child(6) a")
+            if link:
+                # Links look like this:
+                # javascript:__doPostBack('<eventtarget>','<eventargument>')
+                eventtarget, eventargument = link.css("::attr(href)")\
+                                                 .re(r"'(.*?)'")
+                anchor = response.meta['source_anchor']
+                anchor += " ".join(row.css("td::text").extract()[:3])
+                anchor += " " + link.css("::text").extract_first()
```
HarrisonGregg: I try to avoid using strings as accumulators in Python on principle, since they're immutable.  How about something like this:
```python
                anchor = " ".join([
                    response.meta['department_name'],
                    *row.css("td::text").extract()[:3],
                    link.css("::text").extract_first()
                ])
```

osp_scraper/spiders/uga.py
jodizzle: Using the name `source_anchor` for anchor pieces I carry over to the next method is something I think I've done in several scrapers now, but I think you're right.  Using a different name is probably better.

osp_scraper/spiders/uga.py
jodizzle: The reason I do this is for consistency with scrapers that do implement `extract_links`.  Is that not valuable?  Is there a reason we would pull both for a scraper that implemented `extract_links`?

osp_scraper/spiders/uga.py
jodizzle: Oops, that's really bad.  A hardcoded value like that would tend to cause errors.

What's weird is that the errors only come up following some requests, sometimes.  There doesn't seem to be too much consistency.

osp_scraper/spiders/uga.py
jodizzle: Technically I use this when grabbing the search result page, but that decision is in dispute.

osp_scraper/spiders/uga.py
jodizzle: Actually, it being `source_anchor` is currently required for grabbing the search result page, but that's in dispute.

jodizzle:
@HarrisonGregg I made some changes.

osp_scraper/spiders/uga.py
HarrisonGregg: The only reason we save search pages as WARC files is because of the way that the files pipeline works, not because we actually need the search page.  I don't know much about osp-pipeline, but my impression is that it has no use for a search results page. It's just looking for syllabus pages.

osp_scraper/spiders/uga.py
HarrisonGregg: Oh right.  I forgot about that. But yeah, there's no reason to call `parse_for_files` on the search results pages, so you can change these too.

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,83 @@
...
+                # the lowest amount of form-work.
+                'RadioButtonList1': "DEP1",
+                'Button1': "Submit"
+            },
+            callback=self.parse_for_departments
+        )
+
+    def parse_for_departments(self, response):
+        departments = response.css("#ddlDepartment option")[1:]
+        for dept in departments[:7]:
```
HarrisonGregg: Is this test code?

osp_scraper/spiders/uga.py
```diff
@@ -0,0 +1,83 @@
...
+    def parse_for_courses(self, response):
+        # Grab the search result page.
+        for item in self.parse_for_files(response):
+            yield item
+
+        # Syllabi are stored at the same URL behind javascript links that feed
+        # into ASP.NET POST requests.  So, rather than implement extract_links,
+        # we yield the response of each POST request into parse_for_files.
+        rows = response.css("#gridFileList tr")[1:]
+        for row in rows[:3]:
```
HarrisonGregg: Test code?

jodizzle:
@HarrisonGregg I removed the scrape of the search page and removed the test code (oops).

PR #83, opened 2017-05-05
Fixup PDF scraper

jodizzle:
Changing the PDF scraper to use `database_urls` is required per 0fe7ca1.  In addition, there were latent issues with the PDF scraper not setting meta parameters.

### DIFF
--- a/osp_scraper/spiders/pdf.py
+++ b/osp_scraper/spiders/pdf.py
@@ -25,12 +25,8 @@ class PDFSpider(CustomSpider):
                     'source_url': start_url,
                     'source_anchor': start_url
                 },
-                callback=self.parse
+                callback=self.parse_for_files
             )
-
-    def parse(self, response):
-        for item in self.parse_for_files(response):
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/pdf.py
```diff
@@ -12,9 +13,21 @@ class PDFSpider(CustomSpider):
...
+        for start_url in self.database_urls:
+            yield scrapy.Request(
+                start_url,
+                meta={
+                    'depth': 0,
+                    'hops_from_seed': 0,
+                    'source_url': start_url,
+                    'source_anchor': start_url
+                },
+                callback=self.parse
```
HarrisonGregg: Can you change this to `self.parse_for_files` and remove the `parse` function below?

jodizzle: Oh, I didn't even notice it was there.

jodizzle:
@HarrisonGregg

PR #84, opened 2017-05-05
edurepocrawler: Check for 'Custom Scraper Name' properly

jodizzle:
The rows extracted by `csv.DictReader` are populated such that `'Custom Scraper Name' in row` will return true for every row that has a 'Custom Scraper Name' cell (i.e., every row), even if it's empty. This caused errors when trying to run the `crawl_func` for each custom scraper.

Sidenote: Is there a reason for the `not local` part of the `if` statement?  Why wouldn't you want to run custom scrapers while running locally?

### DIFF
--- a/bin/edu_repo_crawler.py
+++ b/bin/edu_repo_crawler.py
@@ -27,7 +27,7 @@ def main(csv_file, local, institution):
         for row in csv.DictReader(f):
             if not institution or institution == row['id']:
                 # Run custom scraper, but only if not running locally.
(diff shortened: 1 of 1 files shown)


jodizzle:
I usually take the `local` out of the `if` statement when testing the script, and I didn't notice any Twisted errors.  However, that's likely because I combine `--local` and `--institution` to test a specific row.

There should probably be some further small re-configuring of the logic in this script, but it's not a pressing issue for now.  I'll merge this PR.

PR #86, opened 2017-05-06
Add Snow College scraper

jodizzle:
https://www.snow.edu/syllabus/?action=search&subject=normal

My approach here is a bit messy, but it was cleanest I could come up with.  It might be worth a second look.

### DIFF
--- a/osp_scraper/spiders/snow.py
+++ b/osp_scraper/spiders/snow.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SnowSpider(CustomSpider):
+    name = "snow"
+
+    start_urls = ["https://www.snow.edu/syllabus/?action=search&subject=normal"]
+
+    def start_requests(self):
+        for request in super().start_requests():
+            request.meta['source_url'] = request.url
+            request.meta['source_anchor'] = '1'
+            yield request
+
+    def parse(self, response):
+        for item in self.parse_for_files(response):
+            yield item
+
+        page_tags = response.css("table tr:nth-child(2) td a")
+        # Check to see if there's a 'Next' anywhere in the tags list.
+        if page_tags.re_first(r"Next"):
+            # The 'href' attribute of each link looks like:
+            # javascript:searchPage(<page number>)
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/snow.py
```diff
@@ -0,0 +1,63 @@
...
+            request.meta['source_anchor'] = '1'
+            yield request
+
+    def parse(self, response):
+        for item in self.parse_for_files(response):
+            yield item
+
+        page_tags = response.css("table tr:nth-child(2) td a")
+        # Check to see if there's a 'Next' anywhere in the tags list.
+        if 'Next' in ''.join(page_tags.css("::text").extract()):
```
HarrisonGregg: Would this work?
```python
if page_tags.re_first(r"Next"):
```

HarrisonGregg: Alternatively, something like this streamlines it a bit:
```python
next_page_num = response.css("table tr:nth-child(2) td a")\
    .re_first(r'"javascript:searchPage\((\d+)\)">Next')
if next_page_num:
```

HarrisonGregg: Or, you could refactor to let scrapy parallelize it a bit more by generating the requests for all the pages at the beginning. I think that's the more scrapy way to do it.
```python
last_page_num = response.css("table tr:nth-child(2) td a")\
    .re(r'"javascript:searchPage\((\d+)\)">\d')[-1]
for page_num in range(1, int(last_page_num)+1):
```

jodizzle: Your last example is intriguing.  Would that actually let Scrapy parallelize better?  I'm not sure how much of a distinction there is between an iterative approach and a recursive approach, but I don't know much about Scrapy internals.

For now I'll implement your first example, which I think is very clean.

jodizzle:
@HarrisonGregg I also used a more sensible source for `source_url` in `start_requests`.

osp_scraper/spiders/snow.py
HarrisonGregg: I think it would make more of a difference if we didn't set `download_delay` so high, but it would make a difference if one request is taking longer than `download_delay` for some reason, or a request returns and error.  Doing it that way could also let you clean up the scraper by removing `start_requests` and the direct call to `parse_for_files`.

PR #87, opened 2017-05-06
Add South Dakota School of Mines and Technology scraper

jodizzle:
https://sdmines.sdsmt.edu/cgi-bin/global/tech_search_dir_syllabus.cgi

Fairly straightforward when using course abbreviations from [here](http://ecatalog.sdsmt.edu/content.php?catoid=14&navoid=2598).

### DIFF
--- a/osp_scraper/spiders/sdsmt.py
+++ b/osp_scraper/spiders/sdsmt.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SDSMTSpider(CustomSpider):
+    """
+    The search for this site can be done by course abbreviations, which are
+    available in a table on a separate page.
+    """
+    name = "sdsmt"
+
+    start_urls = ["http://ecatalog.sdsmt.edu/content.php?catoid=14&navoid=2598"]
+
+    def parse(self, response):
+        search_url = "https://sdmines.sdsmt.edu/cgi-bin/global/tech_search_dir_syllabus.cgi"
+
+        rows = response.css("td.block_content table table tr")
+        for row in rows[2:]:
+            abbrev = self.clean_whitespace(
+                        row.css("td:first-child::text").extract_first())
+            anchor = row.css("td:nth-child(2)::text").extract_first()
+            yield scrapy.FormRequest(
+                search_url,
+                method="POST",
+                formdata={
+                    'submitbtn': "Search",
+                    'search': abbrev,
+                    'termtype': "all"
+                },
+                meta={
+                    'depth': 1,
+                    'hops_from_seed': 1,
+                    'source_url': response.url,
+                    'source_anchor': anchor
+                },
+                callback=self.parse_for_files
+            )
+
+    def extract_links(self, response):
+        rows = response.css("table.table tr")
+        for row in rows[1:]:
+            rel_url = row.css("td:nth-child(4) a::attr(href)").extract_first()
+            url = response.urljoin(rel_url)
+            anchor = " ".join(row.css("::text").extract())
+
+            yield (url, anchor)


osp_scraper/spiders/sdsmt.py
```diff
@@ -0,0 +1,49 @@
...
+    available in a table on a separate page.
+    """
+    name = "sdsmt"
+
+    start_urls = ["http://ecatalog.sdsmt.edu/content.php?catoid=14&navoid=2598"]
+
+    def parse(self, response):
+        search_url = "https://sdmines.sdsmt.edu/cgi-bin/global/tech_search_dir_syllabus.cgi"
+
+        rows = response.css("td.block_content_outer table table tr:nth-child(3) table tr")
```
HarrisonGregg: Looks like you can get away with just:
```python
        rows = response.css(".block_content table table tr")
```

osp_scraper/spiders/sdsmt.py
```diff
@@ -0,0 +1,49 @@
...
+        for row in rows[2:]:
+            abbrev = self.clean_whitespace(
+                        row.css("td:first-child::text").extract_first())
```
HarrisonGregg: Just because I don't like the way the line breaks here, you could do it like this instead:
```python
            abbrev = row.css("td:first-child::text").re_first(r"[A-Z]+")
```

osp_scraper/spiders/sdsmt.py
```diff
@@ -0,0 +1,49 @@
...
+                callback=self.parse_for_files
+            )
+
+
+    def extract_links(self, response):
+        rows = response.css("table.table tr")
+        for row in rows[1:]:
+            rel_url = row.css("td:nth-child(4) a::attr(href)").extract_first()
+            url = response.urljoin(rel_url)
+            anchor = ' '.join(row.css(" ::text").extract())
```
HarrisonGregg: Minor: extra space in `" ::text"`.

HarrisonGregg: Also minor, but for consistency, quotes should be double.

osp_scraper/spiders/sdsmt.py
jodizzle: Yeah, this is better.  I'm not sure how I arrived at the one that's currently there, anyway.

osp_scraper/spiders/sdsmt.py
jodizzle: I'm not sure I like this.  While it would currently work, and would probably work into the future, extracting the exact string from the cell feels more secure to me than relying on a pattern match.  Unless there's a nice way to extract whatever is in the cell without having to run `self.clean_whitespace` over it?

osp_scraper/spiders/sdsmt.py
jodizzle: But I'm using  `td.block_content` instead of `.block_content` on principle, since classes aren't necessarily intended to be unique in HTML.

osp_scraper/spiders/sdsmt.py
HarrisonGregg: Really? I feel like it's more secure to specify explicitly that you're looking for a series of capital letters, given that you know the course abbreviation is going to be in that form.  In all likelihood, both will work indefinitely into the future, but in what way is it more secure to use whatever you get from the cell instead of specifying the pattern you're looking for?

jodizzle: What I meant is that it's possible, theoretically, that the course abbreviation might not just be capital letters, and in that sense it's more "secure" to just take what's in the cell (e.g., exactly what the institution specifies) and clean it up rather than just assume that it's going to be capital letters.  But maybe that's a minor concern and/or there's a more general regex that could be used.

osp_scraper/spiders/sdsmt.py
jodizzle: The extra space was actually intentional, but it looks like it's not necessary.

osp_scraper/spiders/sdsmt.py
HarrisonGregg: Part of what bothered me was also that the HTML has escape code `&#160;`, which I guess scrapy automatically unescapes into a non breaking space.

You could also just do:
```python
            abbrev = row.css("td:first-child::text").strip()
```
since `clean_whitespace` is mostly for when we join a bunch of `::text`s together that might have extra white space in the middle, but you can leave it however you want.

PR #89, opened 2017-05-07
Add Texas State University scraper

jodizzle:
https://api.hb2504.txstate.edu/py/getdepartments.py

This is one of those ones with a nice API.

### DIFF
--- a/osp_scraper/spiders/txstate.py
+++ b/osp_scraper/spiders/txstate.py
@@ -1,0 +1,95 @@
+# -*- coding: utf-8 -*-
+
+import json
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class TXStateSpider(CustomSpider):
+    name = "txstate"
+
+    start_urls = ["https://api.hb2504.txstate.edu/py/getdepartments.py"]
+
+    def parse(self, response):
+        departments_url = "https://api.hb2504.txstate.edu/py/getcoursesfordepartment.py"
+        jsonresponse = json.loads(response.body_as_unicode())
+        for dept in jsonresponse['departments']:
+            yield scrapy.FormRequest(
+                departments_url,
+                method="GET",
+                formdata={
+                    'department': dept
+                },
+                callback=self.parse_for_courses
+            )
+
+    def parse_for_courses(self, response):
+        courses_url = "https://api.hb2504.txstate.edu/py/getcoursedetails.py"
+        jsonresponse = json.loads(response.body_as_unicode())
+        for course in jsonresponse['courses']:
+            number = course['number']
+            title = course['title']
+            yield scrapy.FormRequest(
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/txstate.py
```diff
@@ -0,0 +1,91 @@
...
+        jsonresponse = json.loads(response.body_as_unicode())
+        for row in jsonresponse['classes']:
+            # The json provides a convenient boolean for whether a class has a
+            # syllabus available or not.
+            if row['syllabus']:
+                indexnum = row['indexnum']
+                anchor_terms = dict(
+                        response.meta,
+                        section=row['section'],
+                        displayname=row['instructor']['displayname'])
```
HarrisonGregg: For consistency, this should be indented only once and the closing parenthesis should be on a new line.

osp_scraper/spiders/txstate.py
```diff
@@ -0,0 +1,91 @@
...
+                anchor = "{course} {section} {title} {semester} {displayname}".format(**anchor_terms)
```
HarrisonGregg: It seems like it's not quite worth the added complexity of doing it this way just to avoid pulling the elements out of `meta` individually.  It might be clearer to do something like this:
```python
                anchor = " ".join([
                    response.meta['course'],
                    row['section'],
                    response.meta['title'],
                    response.meta['semester'],
                    row['instructor']['displayname']
                ])
```

HarrisonGregg: Or if you do want to avoid pulling them out individually, it seems unnecessary to pack the arguments to a function into a dict, so something like this might be okay too:
```python
                anchor = "{course} {section} {title} {semester} {instructor}".format(
                    **response.meta,
                    section=row['section'],
                    instructor=row['instructor']['displayname']
                )
```

osp_scraper/spiders/txstate.py
```diff
@@ -0,0 +1,91 @@
+# -*- coding: utf-8 -*-
+
+import json
+import scrapy
```
HarrisonGregg: We haven't been consistent on this, but I think we should separate standard library imports and third party imports with an extra newline.

jodizzle: Yep, my bad.

osp_scraper/spiders/txstate.py
jodizzle: Oh sure.  I had been playing around with leaving the closing parenthesis on the last line, but I guess it makes more sense to put it on a newline to be consistent with how we format the Scrapy request calls.

osp_scraper/spiders/txstate.py
jodizzle: Maybe I got a bit too fancy with my string formatting.  I'll use your first example.

PR #90, opened 2017-05-07
Add College of Marin scraper

jodizzle:
http://www.marin.edu/Apps/Directory/CourseSearch.aspx?Letter=A

Straightforward.

### DIFF
--- a/osp_scraper/spiders/marin.py
+++ b/osp_scraper/spiders/marin.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class MarinSpider(CustomSpider):
+    name = "marin"
+
+    start_urls = ["http://www.marin.edu/Apps/Directory/CourseSearch.aspx"]
+
+    def parse(self, response):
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/marin.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class MarinSpider(CustomSpider):
+    name = "marin"
+
+    start_urls = ["http://www.marin.edu/Apps/Directory/CourseSearch.aspx?Letter=A"]
```
HarrisonGregg: I don't think it's worth this much added complexity to avoid requesting the first page twice.  Personally, I'd remove `"?Letter=A"` from here, remove `start_requests`, remove lines 19-21, and remove the splice on 23, and everything should still work.

jodizzle: I guess my current approach is heavy-handed for just avoiding one request.  I'll refactor.

jodizzle:
@HarrisonGregg Seems like your strategy works.  Merging.

PR #92, opened 2017-05-08
Follow frame/iframe src URLs as links in FilterSpider

HarrisonGregg:
I thought this seemed reasonable, as frames/iframes in HTML are as valuable or more valuable to follow than `a` tag links.  There are a few sites that we've found that weren't getting scraped properly because links were embedded in frames.  On an individual basis, we can change the URL to that of the frame, but this should fix it more generally and shouldn't have any significant downsides.  @wearpants Thoughts?

### DIFF
--- a/osp_scraper/spiders/__init__.py
+++ b/osp_scraper/spiders/__init__.py
@@ -134,6 +134,15 @@ class FilterSpider(BaseSpider):
 
             yield (url, anchor)
 
+        for frame in (response.css("frame") + response.css("iframe")):
+            relative_url = frame.css("::attr(src)").extract_first()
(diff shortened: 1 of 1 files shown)


HarrisonGregg:
An example is [http://www.gavilan.edu/curriculum/frame_co.html](http://www.gavilan.edu/curriculum/frame_co.html). It seems to work for that.

HarrisonGregg:
@wearpants You said you approved of the idea. Did you want to take a quick look at the implementation before I merge it?

HarrisonGregg:
@wearpants You said you approved of the idea. Did you want to take a quick look at the implementation before I merge it?

Issue #97, opened 2017-05-23
Take advantage of new Scrapy 1.4.0 features


jodizzle:
With https://github.com/opensyllabus/osp-scraper/pull/96, we have access to several new features that could be of use to us.  Here are a few:

- [`response.follow`](https://scrapy.readthedocs.io/en/latest/topics/request-response.html#scrapy.http.TextResponse.follow) for `Request`s.  See the example [here](https://scrapy.readthedocs.io/en/latest/news.html#scrapy-1-4-0-2017-05-18)
- Dropping of form fields in `FormRequest.from_response`, as per [this issue](https://github.com/scrapy/scrapy/issues/667).  Useful for some of our custom scrapers.
- `SelectorList.get` and `SelectorList.getall`.  See [here](https://parsel.readthedocs.io/en/latest/history.html#id1).
- [Link extractors](https://doc.scrapy.org/en/latest/topics/link-extractors.html#link-extractors).  I don't think this is exactly a new feature, but there were several updates to Link Extractors in Scrapy 1.4.0.

There's probably lots more.

PR #98, opened 2017-05-23
Refactor UTSA scraper to use scrapy 1.4 features

HarrisonGregg:


osp_scraper/spiders/utsa.py
```diff
@@ -13,46 +13,20 @@ def parse(self, response):
         subjs = response\
             .css("#ctl00_MainContentSearchQuery_searchCriteriaEntry_CourseSubjectCombo_OptionList ")\
             .css("li::text")\
-            .extract()
+            .getall()
```
jodizzle: - Do we want to standardize `getall` instead of `extract` and `get` instead of `extract_first`?  I mentioned it in [my issue](https://github.com/opensyllabus/osp-scraper/issues/97) but I'm not too attached.
- Why is this call spread across three lines?  Just consistency?

HarrisonGregg: - `getall` vs. `extract` doesn't make that much of difference in terms of line length, but given that `get` is a lot shorter than `extract_first`, I think we should use it, and so we should use `getall` for consistency.  
 - You were the one that originally wrote it that way, but I think it makes sense.  Stylistically, I think it's easier to read if you put one function call on each line if it breaks at all.

-

jodizzle: Oh, I think I remember now.  I guess PRs can be a time for introspection as well...

PR #99, opened 2017-05-23
Add functional Diablo Valley College scraper

HarrisonGregg:
Pretty straightforward.


osp_scraper/spiders/dvc.py
```diff
@@ -0,0 +1,41 @@
...
+            yield scrapy.FormRequest.from_response(
+                response,
+                formdata={
+                    'subject': subject_value
+                },
+                callback=self.parse_for_courses
+            )
+
+    def parse_for_courses(self, response):
+        table = response.css("table[width='85%']")
```
jodizzle: I suppose it's not that big of an issue, but I find using `width` or similar types of attributes (e.g., `height`) as part of selectors to be a bit strange.  Could you use `:first-child` or `:first-of-type` somehow instead?

HarrisonGregg: I know what you mean, and I've felt the same way, but I'm not sure I agree.  If there are no names, classes, or ids, then however we end up identifying it is kind of arbitrary.  I could do something like `p + table`, but that's equally arbitrary.  If anything about the page changes, it could break anything, so I think it makes more sense to identify the table by its own property instead of it's relative position to other elements.

jodizzle: I still don't agree, but I don't think I have many concrete arguments against what you're saying, so we can leave it.

(The best I have is that while it is true that any changes on a page could break the custom scraper, some parts of the page might be more likely to change than others.  The `width` element is one of those parts.)

HarrisonGregg: It looks like the other table on the page has hardly anything in it, so just selecting `table` and then looking for `b` and `a` tags within a table gets us what we're looking for.  This could still lead to errors if the page structure was changed, but this approach seems the post likely to keep downloading the pages we want, even if there are some errors.

jodizzle: That seems fine.

osp_scraper/spiders/dvc.py
```diff
@@ -0,0 +1,41 @@
...
+from ..spiders.CustomSpider import CustomSpider
+
+class DVCSpider(CustomSpider):
+    name = "dvc"
+
+    start_urls = ["http://www3.dvc.edu/org/info/schedules/course-outline.htm"]
+
+    def parse(self, response):
+        options = response.css("#subject option")[1:]
+        for option in options:
```
jodizzle: Should the variable name be something more descriptive than `options`?  I think `subjects` would be most similar to the type of naming we usually do.

Also, I realized that this site allows you to do an empty search, which I think (?) returns a list of all courses.  So maybe you don't need to loop at all?

HarrisonGregg: Turns out a GET request to [http://www3.dvc.edu/org/info/schedules/course-outline-results.htm](http://www3.dvc.edu/org/info/schedules/course-outline-results.htm) will return all of the results, so this scraper isn't necessary.

HarrisonGregg:
Turns out a GET request to [http://www3.dvc.edu/org/info/schedules/course-outline-results.htm](http://www3.dvc.edu/org/info/schedules/course-outline-results.htm) will return all of the results, so this scraper isn't necessary.

PR #101, opened 2017-05-24
Add functional Tompkins Cortland Scraper

HarrisonGregg:


osp_scraper/spiders/tompkinscortland.py
```diff
@@ -0,0 +1,50 @@
...
+                    meta={
+                        'depth': 1,
+                        'hops_from_seed': 1,
+                        'source_url': response.url,
+                        'source_anchor': course_name,
+                    },
+                    callback=self.parse_for_files
+                )
+
+        pages = response.css("#trTableFooterPager a::text").getall()
```
jodizzle: I believe `trTableFooterPager` is a `class`, not an `id`.

osp_scraper/spiders/tompkinscortland.py
```diff
@@ -0,0 +1,50 @@
...
+        for page in pages:
+            yield scrapy.FormRequest.from_response(
```
jodizzle: Couldn't this logic create multiple requests for each page?  For example, if a `response` contains two pages, then on page one, this loop will iterate over `['2']` and make a request with a `callback` to `self.parse_for_courses`.  But on page two, this loop will iterate over `['1']` and make an unnecessary request to page one again.  The problem gets worse the more pages there are.

I suppose you could rely on Scrapy's filtering to handle duplicate requests, but it seems messy to me.

osp_scraper/spiders/tompkinscortland.py
```diff
@@ -0,0 +1,50 @@
...
+                    '__EVENTTARGET': "ctl00$mainContent$gvCourses",
+                    '__EVENTARGUMENT': "Page$" + page
+                },
+                callback=self.parse_for_courses
+            )
+
+    def extract_links(self, response):
+        for a_tag in response.css("a"):
+            url = a_tag.css("::attr(href)").get()
+            if url and "syllabus" in url.lower():
```
jodizzle: Does this type of thing actually select the syllabi?  The term `syllabus` doesn't seem to be contained in any of the syllabus links I can find.

HarrisonGregg: Oops. I meant anchor.

osp_scraper/spiders/tompkinscortland.py
HarrisonGregg: Good catch.

osp_scraper/spiders/tompkinscortland.py
HarrisonGregg: Good catch.

HarrisonGregg:
@jodizzle Guess I was moving a little too quickly with this one.  I've done some more rigorous testing and it seems to work now.

osp_scraper/spiders/tompkinscortland.py
```diff
@@ -0,0 +1,55 @@
...
+                    },
+                    callback=self.parse_for_files
+                )
+
+    def extract_links(self, response):
+        course_name = response.css(".catCourseName::text").get()
+        for a_tag in response.css("a"):
+            url = a_tag.css("::attr(href)").get()
+            anchor = " ".join(a_tag.css("::text").getall())
+            if url and "syllabus" in anchor.lower():
```
jodizzle: Is there a reason you check based on the `anchor` and not the `url`?  The URL always seems to contain 'MCSdisplay.asp' (MCS stands for Master Course Syllabus, I assume).  If you checked based on the URL, you could also move anchor creation inside the `if` statement.

Also, some syllabi links seem to resolve to "Not Found" pages, like [this one](http://www.tompkinscortland.edu/mcs/MCSdisplay.asp?cid=BIOL+NOLAB).  Unfortunately, they have 200 status codes, so I'm not sure if there's anything we can (cleanly) do about them.

HarrisonGregg: My thinking was just that in case there are any pages that have syllabus links that are different in any way, this should cover that.

jodizzle: Sure, I guess that's fine.

HarrisonGregg: Not sure what to do about the "Syllabi Not Found" pages, given that they don't even redirect or anything.  I'd have to rewrite the scraper and check each page.  It seems like there aren't that many, so I don't think it's necessary.

PR #102, opened 2017-05-24
Add functional University of Texas Arlington scraper

HarrisonGregg:


osp_scraper/spiders/uta.py
```diff
@@ -0,0 +1,54 @@
...
+                },
+                callback=self.parse_for_courses
+            )
+
+    def parse_for_courses(self, response):
+        html = json.loads(response.body)["courseResultCards"]
+        sel = scrapy.Selector(text=html, type="html")
+        courses = sel.css(".result-card")
+        for course in courses:
+            url = course.css(".syllabus::attr(href)").get()
```
jodizzle: I don't think it's that much of a concern, but in my tests, it seemed like the first few elements of `courses` were duds, and went to http://mentis.uta.edu/dashboard/file/download/id/.

osp_scraper/spiders/uta.py
```diff
@@ -0,0 +1,54 @@
...
+    All of the syllabus files are behind redirects, so this scraper requests
+    the file directly instead of using the file pipeline.
+    """
+    name = "uta"
+
+    start_urls = ["https://mentis.uta.edu/explore/courses/all"]
+
+    def parse(self, response):
+        num_results = int(response.css("#search-message strong::text").get())
+        num_pages = (num_results+9)//10
```
HarrisonGregg: All it's doing is dividing by 10 (the page size) and rounding up, but I can add a comment.

Why would `int` go around `num_pages` instead? I need `num_results` to be a number so I can divide it, and the `//` ensures that `num_pages` will be an integer.

osp_scraper/spiders/uta.py
HarrisonGregg: Good catch. I added a check to avoid the unnecessary requests.

osp_scraper/spiders/uta.py
jodizzle: Oh, my bad.  I didn't see the `//` there.  I thought it was only one `/`.

PR #103, opened 2017-05-24
Add functional Fort Hays State University scraper

HarrisonGregg:
This scraper naively scrapes all URLs that might lead to a file.

### DIFF
--- a/osp_scraper/spiders/fhsu.py
+++ b/osp_scraper/spiders/fhsu.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class FHSUSpider(CustomSpider):
+    """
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/fhsu.py
```diff
@@ -0,0 +1,34 @@
...
+    are of the form seen below.  It was non-trivial to implement searches
+    through the form itself, so instead, this scraper grabs the files at all
+    urls of this form.  Currently, the URLs lead to files up to id=593.  The
+    scraper looks for files up to id=1000 for some amount of future proofing.
+    """
+
+    name = "fhsu"
+
+    custom_settings = {
+        'RETRY_ENABLED': False
```
jodizzle: This is to prevent excessive retries on file ids above 593, right?

HarrisonGregg: That's the idea.  For some reason, the server returns 500 instead of 400 errors, so it would normally retry twice, which seems unnecessary.

PR #104, opened 2017-05-25
Add 'rtf' to ALLOWED_FILE_TYPES

jodizzle:
With the addition of [pipeline support for RTF files](https://github.com/opensyllabus/osp-pipeline/pull/34), we should be allowing RTF file types here as well.

### DIFF
--- a/osp_scraper/spiders/__init__.py
+++ b/osp_scraper/spiders/__init__.py
@@ -13,7 +13,7 @@ from .. import version
 from ..filterware import Filter
 
 # file types we download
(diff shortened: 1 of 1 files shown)


HarrisonGregg:
@jodizzle Technically, I don't think this is necessary, since `ALLOWED_FILE_TYPES` is only currently used to filter binary files, and RTFs are text, not binary.

jodizzle:
@HarrisonGregg Oops, my bad.  That's good to know, I misunderstood how the `ALLOWED_FILE_TYPES` was being evoked.  Let's call this "future proofing".

PR #105, opened 2017-05-28
Add functional Texas Tech University scraper

HarrisonGregg:


osp_scraper/spiders/ttu.py
```diff
@@ -0,0 +1,96 @@
...
+            .getall()[1:]
+        for prefix in prefixes:
+            # convert prefix to hexadecimal ASCII representation
+            hex_prefix = "".join(format(ord(c), "X") for c in prefix)
+            yield scrapy.FormRequest(
+                "http://appserv.itts.ttu.edu/PACI/Pages/Search/SearchResults.aspx",
+                method="GET",
+                formdata={
+                    '7374797065': "434F55525345",
+                    '63707265666978': hex_prefix,
```
HarrisonGregg: I looked at them, and this seemed like the best option.

osp_scraper/spiders/ttu.py
```diff
@@ -0,0 +1,96 @@
...
+
+from ..spiders.CustomSpider import CustomSpider
+
+class TTUSpider(CustomSpider):
+    name = "ttu"
+
+    start_urls = ["http://catalog.ttu.edu/content.php?catoid=2&navoid=163"]
+
+    custom_settings = {
+        'ROBOTSTXT_OBEY': False,
```
HarrisonGregg: This overrides the entire `CUSTOM_SETTINGS` property, so it doesn't inherit it.

osp_scraper/spiders/ttu.py
```diff
@@ -0,0 +1,96 @@
...
+        'USER_AGENT': "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.81 Safari/537.36",
```
HarrisonGregg: I tried something like that but it didn't seem to work.

osp_scraper/spiders/ttu.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def parse_for_courses(self, response):
+        for item in self.parse_for_files(response):
+            yield item
+
+        page = response.meta["page"] + 1
+        prefix = response.meta["prefix"]
+
+        # __EVENTVALIDATION and __VIEWSTATE are included in a form in the first
+        # page response, but they sometimes change when the page changes, and
```
HarrisonGregg: They're always included, but they only change sometimes.

osp_scraper/spiders/ttu.py
```diff
@@ -0,0 +1,96 @@
...
+        yield scrapy.FormRequest.from_response(
+            first_response,
+            headers={
+                'Accept': "*/*",
+                'X-MicrosoftAjax': 'Delta=true'
+            },
+            formdata={
+                '__ASYNCPOST': "true",
+                '__LASTFOCUS': "",
+                'ctl00$ContentPlaceHolder1$ddl_CoursePageSize': "10",
```
HarrisonGregg: Yeah, I tried, but couldn't get it to work properly. One issue I ran into was that if I requested a page that was past the number of results with page size 100, it would start returning the pages of size 10.

osp_scraper/spiders/ttu.py
jodizzle: Oh I see.  Good to know.

Do you think it would be appropriate to do something like
```python
custom_settings = {
    **super().custom_settings,
    'USER_AGENT': #etc.
}
```
?

osp_scraper/spiders/ttu.py
jodizzle: When I do a test run over a couple prefixes and a couple pages, I get the same number of downloads whether I use your `USER_AGENT` or just `USER_AGENT: "Mozilla/5.0"`.  Do you remember where it wasn't working in your tests?

Interestingly, when I run it without a `USER_AGENT`, I don't get errors or anything, the yield is just much lower.

osp_scraper/spiders/ttu.py
jodizzle: Ah, that's unfortunate.  Probably not worth it then.

osp_scraper/spiders/ttu.py
jodizzle: Alright, your comment is fine then.

osp_scraper/spiders/ttu.py
HarrisonGregg: Weird. When I was testing it, it definitely seemed like the requests would just time out if I didn't include a user agent, but testing it now, it seems like they work just fine without it.

osp_scraper/spiders/ttu.py
HarrisonGregg: Hmm, looking into it further, it does look like the syllabus search would be preferable, since you can search with no input and get all the syllabi, but the paging is done differently, and I couldn't get it to work.

HarrisonGregg:
@jodizzle I removed the user agent, as it doesn't seem necessary after all, but I don't think there's anything else I can do to improve it.

PR #43, opened 2017-03-20
Added functional Sheridan College scraper

HarrisonGregg:

PR #106, opened 2017-05-30
Add functional Wesleyan scraper

HarrisonGregg:
Straightforward scraper to use the public API.

### DIFF
--- a/osp_scraper/spiders/wesleyan.py
+++ b/osp_scraper/spiders/wesleyan.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+import datetime
+import json
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class WesleyanSpider(CustomSpider):
+    name = "wesleyan"
+
+    def start_requests(self):
+        # Terms are generated with JS on page, so logic is replicated here.
+        terms = []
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/wesleyan.py
```diff
@@ -0,0 +1,52 @@
...
+
+from ..spiders.CustomSpider import CustomSpider
+
+class WesleyanSpider(CustomSpider):
+    name = "wesleyan"
+
+    def start_requests(self):
+        # Terms are generated with JS on page, so logic is replicated here
+        terms = []
+        current_year = datetime.date.today().year % 100
```
jodizzle: [Be cautious about running on computers with very out of date clocks.](https://github.com/opensyllabus/osp-scraper/pull/43#issuecomment-293644306)

HarrisonGregg: I would have done it the same way I did with that other one, but I didn't think it was worth making an unnecessary request.

HarrisonGregg: Do you have any reason to believe a computer would have an out of date clock?

jodizzle: Oh no, I was mostly just referencing an older PR.  I think using `datetime` is okay, especially in this circumstance.

PR #110, opened 2017-06-05
Add Rio Salado College scraper

jodizzle:
This site is largely ASP.NET javascript shenanigans.  I was able to get it to work, but the result is unfortunately a bit messy.

### DIFF
--- a/osp_scraper/spiders/riosalado.py
+++ b/osp_scraper/spiders/riosalado.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+
+import json
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class RiosaladoSpider(CustomSpider):
+    name = "riosalado"
+
+    start_urls = ["http://www.riosalado.edu/Pages/leftnav/schedule.txt"]
+
+    def parse(self, response):
+        terms = sorted(filter(None, response.css("#terms option::attr(value)").getall()))
+        yield scrapy.Request(
+            "http://www.riosalado.edu/WebServices/Pages/schedule/options.aspx?type=subject",
+            meta={
+                'terms': terms
+            },
+            callback=self.parse_for_subjects
+        )
+
+    def parse_for_subjects(self, response):
+        """
+        This method mimics the logic found in the site's javascript.  When used
+        together, the parameters 'term1' and 'term2' fetch all courses that
+        match a given 'prefix' and occured during 'term1', 'term2' and/or all
+        terms in-between.
+
+        Currently, only two terms are available on the site, but this could
+        change in the future.
+        """
+        subj_url = "http://www.riosalado.edu/WebServices/Pages/schedule/sch.aspx"
+        subjects = json.loads(response.body_as_unicode())
+        term1 = response.meta['terms'][0]
+        term2 = response.meta['terms'][-1]
+        for subj in subjects:
+            yield scrapy.FormRequest(
+                subj_url,
+                method="GET",
+                formdata={
+                    'prefix': subj['Prefix'],
+                    'term1': term1,
+                    'term2': term2
+                },
+                meta={
+                    'title': subj['Subject_area_desc']
+                },
+                callback=self.parse_for_syllabi
+            )
+
+    def parse_for_syllabi(self, response):
+        courses = json.loads(response.body_as_unicode())
+        for course in courses:
+            anchor = " ".join([
+                "Term " + course['Term'],
+                "Section " + course['Sect'],
+                course['Course'],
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/riosalado.py
```diff
@@ -0,0 +1,65 @@
...
+        terms in-between.
+
+        Currently, only two terms are available on the site, but this could
+        change in the future.
+        """
+        subj_url = "http://www.riosalado.edu/WebServices/Pages/schedule/sch.aspx?prefix={0}&term1={1}&term2={2}"
+        subjects = json.loads(response.body_as_unicode())
+        term1 = response.meta['terms'][0]
+        term2 = response.meta['terms'][-1]
+        for subj in subjects[:1]:
```
HarrisonGregg: Is the `[:1]` supposed to be here?

osp_scraper/spiders/riosalado.py
```diff
@@ -0,0 +1,65 @@
...
+            )
+
+    def parse_for_syllabi(self, response):
+        courses = json.loads(response.body_as_unicode())
+        for course in courses:
+            anchor = " ".join([
+                "Term " + course['Term'],
+                "Section " + course['Sect'],
+                course['Course'],
+                course['Title'],
```
HarrisonGregg: Minor: trailing comma.

HarrisonGregg:
I notice that some of the syllabus URLs lead to a page like [this](https://matrix.riosalado.edu/RioWebApps/Public/Rio.SyllabusDefinition.UI/MiniSyllabusRender.aspx?courseId=ACC105&specialtyCode=&instructionMode=IN&startDate=05-01-2017&termCode=4174%C2%A7ionNum=12502).  Do you want to detect those and prevent them from being saved?  I think we've talked about it before and mostly decided against it, so maybe just add a comment about it?

osp_scraper/spiders/riosalado.py
```diff
@@ -0,0 +1,65 @@
...
+
+        Currently, only two terms are available on the site, but this could
+        change in the future.
+        """
+        subj_url = "http://www.riosalado.edu/WebServices/Pages/schedule/sch.aspx?prefix={0}&term1={1}&term2={2}"
+        subjects = json.loads(response.body_as_unicode())
+        term1 = response.meta['terms'][0]
+        term2 = response.meta['terms'][-1]
+        for subj in subjects[:1]:
+            yield scrapy.Request(
```
HarrisonGregg: Is there a reason you're not using a `FormRequest` here?

osp_scraper/spiders/riosalado.py
```diff
@@ -0,0 +1,65 @@
...
+        match a given 'prefix' and occured during 'term1', 'term2' and/or all
+        terms in-between.
+
+        Currently, only two terms are available on the site, but this could
+        change in the future.
+        """
+        subj_url = "http://www.riosalado.edu/WebServices/Pages/schedule/sch.aspx?prefix={0}&term1={1}&term2={2}"
+        subjects = json.loads(response.body_as_unicode())
+        term1 = response.meta['terms'][0]
+        term2 = response.meta['terms'][-1]
```
HarrisonGregg: Alternatively, it looks like you can just use `term1 = 1000` and `term2 = 9999`.  As it is, that seems to get a few courses for 2018, but I'm not sure if any of them have syllabi.

osp_scraper/spiders/riosalado.py
```diff
@@ -0,0 +1,65 @@
...
+
+from ..spiders.CustomSpider import CustomSpider
+
+class RiosaladoSpider(CustomSpider):
+    name = "riosalado"
+
+    start_urls = ["http://www.riosalado.edu/Pages/leftnav/schedule.txt"]
+
+    def parse(self, response):
+        terms = sorted(list(filter(None, response.css("#terms option::attr(value)").getall())))
```
HarrisonGregg: Do you need to include `list`?  Seems like sorted returns a list anyways.

osp_scraper/spiders/riosalado.py
jodizzle: Nope, must be leftover test code.

jodizzle:
@HarrisonGregg I think I addressed the majority of your comments.

As for detecting empty pages, when we've run into those situations in the past we haven't done any extra filtering unless it was convenient and could save us a request.  Does that approach make sense?  Doing extra filtering could save us downloads, but processing in that way seems like more of a job for post-scrape.

As for the terms trick, I think I'd rather deal with what the website gives us, even though my approach places a bit more trust in the website (more pieces functioning as intending) than yours does.  Unless you have strong objections?

PR #111, opened 2017-06-07
Add University of Limerick scraper

jodizzle:
Weird site, but fairly straightforward (if not a bit messy).

### DIFF
--- a/osp_scraper/spiders/ul.py
+++ b/osp_scraper/spiders/ul.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class ULSpider(CustomSpider):
+    '''
+    This site is a bit odd: all search results appear as text embedded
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/ul.py
```diff
@@ -0,0 +1,69 @@
...
+                'source_url': response.url
+            },
+            callback=self.parse_for_pages
+        )
+
+    def parse_for_pages(self, response):
+        # The course title seems to be the most sensible choice of anchor, but
+        # is only available after the page has been loaded.
+        anchor = " ".join(
+            response.css("div.NoticeStyle + div.ModuleLayoutStyle::text").getall()
```
HarrisonGregg: Personally, I'd leave out the `div`s to bring the line under 80 characters.

jodizzle: I usually try to be more specific when using class selectors because classes aren't intended to be unique in HTML.  But I think in this case, you're right.  I'll make the change.

PR #112, opened 2017-06-07
Add Korea Advanced Institute of Science & Technology scraper

jodizzle:
I wrote a lot of comments for this one, but I think they're helpful comments.

### DIFF
--- a/osp_scraper/spiders/kaist.py
+++ b/osp_scraper/spiders/kaist.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+
+import itertools
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class KAISTSpider(CustomSpider):
+    """
+    There is some sort of "session timeout" feature on the search entry page of
+    this site, as well as some sort of queuing feature for when multiple people
+    are accessing the site, both of which could potentially affect the operation
+    of the scraper if run over long periods.
+
+    If a syllabus link is available on a course page, it is scraped; if a
+    syllabus link is not available, the course page itself is scraped.  It seems
+    like most syllabus links contain all the information available on the course
+    pages.
+    """
+    name = "kaist"
+
+    start_urls = ["https://cais.kaist.ac.kr/totalOpeningCourse"]
+
+    def parse(self, response):
+        years = response.css("#sel_year option::attr(value)").getall()
+        terms = response.css("#sel_term option")
+        for year, term in itertools.product(years, terms):
+            term_value = term.css("::attr(value)").get()
+            term_text = term.css("::text").get()
+
+            anchor = " ".join([year, term_text])
+
+            # The department field is populated by a "refresh" action based on
+            # what year and term are selected.
+            yield scrapy.FormRequest(
+                "https://cais.kaist.ac.kr/refreshComboBox",
+                method="POST",
+                formdata={
+                    'year': year,
+                    'term': term_value,
+                    # 'idname' can be other form entry fields, but the one for
+                    # department seems the most useful.
+                    'idname': 'sel_dept',
+                    'lang': 'ENG'
+                },
+                meta={
+                    'year': year,
+                    'term': term_value,
+                    'anchor': anchor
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/kaist.py
```diff
@@ -0,0 +1,166 @@
...
+                        'term': response.meta['term'],
+                        'subject_no': subject_no,
+                        'lecture_class': lecture_class,
+                        'dept_id': response.meta['dept']
+                    },
+                    callback=self.parse_for_syllabi
+                )
+
+    def parse_for_syllabi(self, response):
+        # NOTE: Currently we pull the course page since it seems useful, but
```
HarrisonGregg: The PDFs that I looked at had all of the information contained on the HTML page, so maybe only save the HTML page if there is no syllabus PDF?

jodizzle: Hmm, I think I wanted to avoid making judgements like that, but after looking over it again, I think you're right.

jodizzle:
@HarrisonGregg I followed your suggestion and also fixed a bug where e.g., `row.css(:nth-child(6)::text)` would cause the text of the whole row to be selected if the row was the sixth child, instead of the text of the sixth child of that row, which is intended to be a `td` element.

osp_scraper/spiders/kaist.py
```diff
@@ -0,0 +1,164 @@
...
+                    },
+                    callback=self.parse_for_syllabi
+                )
+
+    def parse_for_syllabi(self, response):
+        # The syllabus link is hidden behind a POST request, so we can't just
+        # use `extract_links`.
+        a_tag = response.css("#fileDownLink")
+        a_value = a_tag.css("::attr(value)").get()
+        if a_value:
```
HarrisonGregg: Huh.  I mentioned _2_ suggestions in my review, but apparently, one didn't make it through.  
It looks like there are some pages like [this one](https://cais.kaist.ac.kr/syllabusInfo?year=2000&term=1&subject_no=73.743&lecture_class=++&dept_id=3406) where `a_value` is the string `"None"`, which causes an error below because there is no `a_text`.

HarrisonGregg: I'd fix it by changing this to:
```python
        a_text = a_tag.css("::text").get()
        if a_value and a_text:
```

osp_scraper/spiders/kaist.py
```diff
@@ -0,0 +1,164 @@
...
+            yield scrapy.FormRequest(
+                "https://cais.kaist.ac.kr/refreshComboBox",
+                method="POST",
+                formdata={
+                    'year': year,
+                    'term': term_value,
+                    # 'idname' can be other form entry fields, but the one for
+                    # department seems the most useful.
+                    'idname': 'sel_dept',
+                    'lang': 'ENG'
```
HarrisonGregg: I think we're trying to scrape all syllabi, not just the English ones

osp_scraper/spiders/kaist.py
jodizzle: Good catch.  Your solution seems reasonable.

osp_scraper/spiders/kaist.py
jodizzle: This is a language selector for the "refresh departments" action, which I'm pretty sure only controls the text of the menus and not the search results.  For the request that executes the actual search, I specify `'sel_lang': 'A'` to select for all languages.

Should I remove `'lang': 'ENG'`?  There seems to be no issue with the request if it's removed from the `formdata`, but then again, it seems good to specify English for the menu text.

jodizzle:
@HarrisonGregg The scraper checks `a_value` and `a_text` now.  In addition, I noticed I was accidentally executing a search for `ALL` departments, which I wanted to specifically avoid by iterating over the departments one-by-one.

osp_scraper/spiders/kaist.py
HarrisonGregg: Ahh, I see.  Leaving it is fine then.

PR #114, opened 2017-06-08
Add Pan-European University scraper

jodizzle:
Pretty straightforward.  It's actually possible to do a search with all years selected, and therefore get all search results in one request, but given how big the response is it seemed safer to break it up a bit.

Also, there's some potential `anchor` content I'm missing out on: There's text adjacent to each syllabus link (outside the `a` tag), but I can't figure out how to (easily) select it.

### DIFF
--- a/osp_scraper/spiders/paneurouni.py
+++ b/osp_scraper/spiders/paneurouni.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class PaneurouniSpider(CustomSpider):
+    name = "paneurouni"
+
+    start_urls = ["https://is.paneurouni.com/katalog/index.pl?jak=dle_pracovist"]
+
+    def parse(self, response):
+        departments = response.css("#fakulty input::attr(value)").getall()
+        years = response.css("#blocks > div")
+        for year in years:
+            year_code = year.css("input[name='obdobi']::attr(value)").get()
+            anchor = year.css("::text").get()
+
+            yield scrapy.FormRequest.from_response(
+                response,
+                method="POST",
+                formdata={
+                    'fakulta': departments,
+                    'obdobi': year_code
+                },
+                meta={
+                    'anchor': anchor
+                },
+                callback=self.do_department_search
+            )
+
+    def do_department_search(self, response):
+        yield scrapy.FormRequest.from_response(
+            response,
+            method="POST",
+            formdata={
+                # The '0' here searches all departments.
+                'ustav': "0",
+                'vypsat': "Display courses"
+            },
+            meta=dict(response.meta),
+            callback=self.parse_for_syllabi
+        )
+
+    def parse_for_syllabi(self, response):
+        a_tags = response.css("a[href^='syllabus']")
+        for a_tag in a_tags:
+            a_tag_text = a_tag.css("::text").get()
+            anchor = response.meta['anchor'] + " " + a_tag_text
+
+            yield response.follow(
+                a_tag,
+                meta={
+                    'depth': 3,
+                    'hops_from_seed': 3,
+                    'source_url': response.url,
+                    'source_anchor': anchor
+                },
+                callback=self.parse_for_files
+            )


osp_scraper/spiders/paneurouni.py
```diff
@@ -0,0 +1,65 @@
...
+
+from ..spiders.CustomSpider import CustomSpider
+
+class PaneurouniSpider(CustomSpider):
+    name = "paneurouni"
+
+    start_urls = ["https://is.paneurouni.com/katalog/index.pl?jak=dle_pracovist"]
+
+    def parse(self, response):
+        fakulta = response.css("#fakulty input::attr(value)").getall()
```
HarrisonGregg: Even if some of the code on the website is in another language, I think all of our variable names should be in English.

osp_scraper/spiders/paneurouni.py
```diff
@@ -0,0 +1,65 @@
...
+        years = response.css("#blocks > div")
+        for year in years:
+            obdobi = year.css("input[name='obdobi']::attr(value)").get()
+            anchor = year.css(" ::text").get()
```
HarrisonGregg: I don't think the space is necessary here.

osp_scraper/spiders/paneurouni.py
```diff
@@ -0,0 +1,65 @@
...
+            },
+            meta={
+                **response.meta
+            },
+            callback=self.parse_for_syllabi
+        )
+
+    def parse_for_syllabi(self, response):
+        a_tags = response.css("a[href^='syllabus']")
+        if a_tags:
```
HarrisonGregg: Is this check necessary? If `a_tags` is empty, iterating through it shouldn't do anything.

osp_scraper/spiders/paneurouni.py
```diff
@@ -0,0 +1,65 @@
...
+            for a_tag in a_tags:
+                a_tag_text = a_tag.css("::text").get()
+                anchor = response.meta['anchor'] + " " + a_tag_text
```
HarrisonGregg: If you really wanted to include the other text in the anchor, you could use regex:
```python
        rows = response.selector.re(r'<a href="syllabus.*?<br>')
        for row in rows:
            sel = scrapy.Selector(text=row)
            a_tag = sel.css("a")[0]

            row_text = " ".join(sel.css("::text").getall())
            anchor = response.meta['anchor'] + " " + row_text
```

osp_scraper/spiders/paneurouni.py
```diff
@@ -0,0 +1,65 @@
...
+    start_urls = ["https://is.paneurouni.com/katalog/index.pl?jak=dle_pracovist"]
+
+    def parse(self, response):
+        fakulta = response.css("#fakulty input::attr(value)").getall()
+        years = response.css("#blocks > div")
+        for year in years:
+            obdobi = year.css("input[name='obdobi']::attr(value)").get()
+            anchor = year.css(" ::text").get()
+
+            yield scrapy.FormRequest(
```
HarrisonGregg: Is there a reason you don't use `from_response` here? Seems like you can use it and omit `vyhledat` and `jak` below.

osp_scraper/spiders/paneurouni.py
```diff
@@ -0,0 +1,65 @@
...
+        yield scrapy.FormRequest.from_response(
+            response,
+            method="POST",
+            formdata={
+                # The '0' here searches all departments.
+                'ustav': "0",
+                'vypsat': "Display courses"
+            },
+            meta={
+                **response.meta
```
HarrisonGregg: I can see why you might want to do it this way for consistency, but I think `dict(response.meta)` or `response.meta.copy()` are preferred for copying dicts in general.

osp_scraper/spiders/paneurouni.py
jodizzle: Really?  I feel like I usually try to have the variable name match up with the HTML elements.  Is that a bad idea?

(That being said, I'm not sure why I chose 'fakulta' instead of 'fakulty' here.  I think it's because the elements under the `#fakulty` tag have `name=fakulta`.)

HarrisonGregg: Yes, definitely.  For understanding the code in the scraper, it's definitely more important to be able to understand what the variable names mean than that the variable names match up with names in the HTML.  When a variable is being used as a value for the key `obdobi` on line 24, having it named `obdobi` is just redundant.  I'd rather have it named `year_code` or something like that.

osp_scraper/spiders/paneurouni.py
jodizzle: Nope, my mistake.

osp_scraper/spiders/paneurouni.py
jodizzle: Sure.

jodizzle: Actually, would it be fine to just use `meta=response.meta` here?

HarrisonGregg: I think that should be avoided, since I think scrapy internals do use `meta` for some other stuff.

osp_scraper/spiders/paneurouni.py
```diff
@@ -0,0 +1,62 @@
...
+        for year in years:
+            year_code = year.css("input[name='obdobi']::attr(value)").get()
+            anchor = year.css("::text").get()
+
+            yield scrapy.FormRequest.from_response(
+                response,
+                method="POST",
+                formdata={
+                    'fakulta': departments,
+                    'obdobi': year_code,
```
HarrisonGregg: Minor: trailing comma.

osp_scraper/spiders/paneurouni.py
jodizzle: Hmm, I suppose that's fair.  I'll use `dict(response.meta)`.

PR #115, opened 2017-06-09
Add City University of Hong Kong scraper

jodizzle:
Pretty straightforward.

### DIFF
--- a/osp_scraper/spiders/cityu.py
+++ b/osp_scraper/spiders/cityu.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+
+from urllib.parse import urljoin
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class CityUSpider(CustomSpider):
+    """
+    This site uses javascript to embed "catalogue pages", which contain all the
+    interesting links, inside the pages which you normally navigate on.
+    However, we can't use the catalogue pages on their own, since all those
+    links are relative links that only work with the URLs of the normal pages,
+    so some maneuvering is required.
+    """
+    name = "cityu"
+
+    start_urls = [
+        "http://www.cityu.edu.hk/pg/201617/catalogue/R/R_course_index.htm"
+    ]
+
+    base_url = "http://www.cityu.edu.hk/pg/201617/catalogue/catalogue_R.htm"
+
+    def parse(self, response):
+        a_tags = response.css("a")
+        for a_tag in a_tags:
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/cityu.py
```diff
@@ -0,0 +1,67 @@
...
+                callback=self.parse_for_syllabi
+            )
+
+    def parse_for_syllabi(self, response):
+        # NOTE: The more natural approach might be to implement `extract_links`,
+        # but that approach downloads both the source page and the link, and
+        # both are syllabi-like, so it might be a good idea to avoid the
+        # possibility of duplication.
+        link = response.css("#pdf_url::text").get()
+        if link:
```
HarrisonGregg: If there isn't a syllabus link, do you think we should save the HTML page?

osp_scraper/spiders/cityu.py
```diff
@@ -0,0 +1,67 @@
...
+    ]
+
+    base_url = "http://www.cityu.edu.hk/pg/201617/catalogue/catalogue_R.htm"
+
+    def parse(self, response):
+        a_tags = response.css("a")
+        for a_tag in a_tags:
+            rel_url = a_tag.css("::attr(href)").get().split("/")[-1]
+            anchor = a_tag.css("::text").get()
+            yield scrapy.FormRequest(
```
HarrisonGregg: Why `FormRequest`? And couldn't you use `response.follow` here instead of `urljoin`?

osp_scraper/spiders/cityu.py
jodizzle: That's a good idea.

jodizzle: Actually, do you have an example of a course page where this happens?  I suppose it wouldn't hurt to throw that case in, but I was having trouble finding an example page to test against.

osp_scraper/spiders/cityu.py
jodizzle: Yep, `FormRequest` must've been leftover from previous approaches.

osp_scraper/spiders/cityu.py
HarrisonGregg: I don't remember, but I guess I didn't actually see any that didn't have a syllabus. I just assumed you had found some because you had the check.

jodizzle: I'll leave it for now but leave an inline comment noting it.

PR #116, opened 2017-06-09
Add Katholische Universität Eichstätt scraper

jodizzle:
This one ended up being more complicated than I expected.

One thing I noticed while testing is that, if I set up a limited crawl across a few semesters and the first few search results, not all semesters (looking at the `anchor`s) are present in the downloads.  I _think_ this is because those results are duplicates of results from other semesters, and Scrapy is doing it's job and filtering duplicate requests.  However, what makes it a bit strange is that setting `HTTPCACHE_ENABLED = True` seems to give different results.  Might just be quirks of Scrapy's behavior.

### DIFF
--- a/osp_scraper/spiders/ku.py
+++ b/osp_scraper/spiders/ku.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class KUSpider(CustomSpider):
+    name = "ku"
+
+    # TODO: Find shorter URL?
+    start_urls = [
+        # The 'IdLanguage=133" means results are in English, or as much English
+        # as is made available -- sometimes of the syllabi are still in German.
+        "https://campus.ku.de/studienangebot/evt_pages/SuchResultat.aspx?node=d175b8fe-7eca-4c19-b50b-7534643c201a&TabKey=WebTab_CstPI2_ModuleAZ&IdLanguage=133"
+    ]
+
+    def parse(self, response):
+        sems = response.css("#ctl00_WebPartManager1_SuchregisterWP1_DropDownList4")\
+            .css("option::attr(value)").getall()
+        for sem in sems[1:]:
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/ku.py
```diff
@@ -0,0 +1,86 @@
...
+                },
+                callback=self.parse_for_pages
+            )
+
+    def parse_for_pages(self, response):
+        for item in self.parse_for_syllabi(response):
+            yield item
+
+        next_page_num = str(response.meta['hops_from_seed'] + 1)
+        eventargument = 'Page$' + next_page_num
```
HarrisonGregg: Minor: should use double quotes for consistency.

osp_scraper/spiders/ku.py
```diff
@@ -0,0 +1,86 @@
...
+
+        yield scrapy.FormRequest.from_response(
+            response,
+            method="POST",
+            formdata={
+                '__EVENTTARGET':
+                    'ctl00$WebPartManager1$ResultatAnzeigenWP1$ctl01',
```
HarrisonGregg: Minor: should use double quotes for consistency.

HarrisonGregg:
I think you're right.  Since we're looking at "modules" instead of courses, there's a lot of overlap between semesters, sometimes 100%.

Is there a reason you're using `HTTPCACHE_ENABLED = True`? I just leave it disabled since it won't be enabled in actual runs and it can cause unexpected behavior sometimes.

osp_scraper/spiders/ku.py
```diff
@@ -0,0 +1,86 @@
...
+                '__EVENTARGUMENT':
+                    eventargument,
```
jodizzle: I think we have some precedent for doing that, but I can ease off on that here.

jodizzle:
I don't usually use `HTTPCACHE_ENABLED = True`.  I just happened to have it enabled for some of my tests for this scraper and it struck me as odd.  But I agree -- testing with cached pages often deviates from non-cached behavior.

PR #118, opened 2017-06-11
Add Stockholm University scraper

jodizzle:
Pretty straightforward thanks to a convenient JSON API.

### DIFF
--- a/osp_scraper/spiders/su.py
+++ b/osp_scraper/spiders/su.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+
+import json
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class SUSpider(CustomSpider):
+    name = "su"
+
+    start_urls = ["https://sisu.it.su.se/en/educations/search.json"]
+
+    def parse(self, response):
+        for item in self.parse_for_urls(response):
+            yield item
+
+        jsonresponse = json.loads(response.body_as_unicode())
+        for page in range(2, jsonresponse['total_pages'] + 1):
+            yield scrapy.FormRequest(
+                self.start_urls[0],
+                method="GET",
+                formdata={
+                    'pg': str(page)
+                },
+                meta={
+                    'depth': response.meta['depth'] + 1,
+                    'hops_from_seed': response.meta['hops_from_seed'] + 1,
+                    'source_anchor': "Page " + str(page)
+                },
+                callback=self.parse_for_urls
+            )
+
+    def parse_for_urls(self, response):
+        jsonresponse = json.loads(response.body_as_unicode())
+        for result in jsonresponse['results']:
+            rel_url = result['url']
+            anchor = " ".join([
+                response.meta.get('source_anchor', "Page 1"),
+                result['code'],
+                result['semester'],
+                result['subject'],
+                result['name']
+            ])
+
+            yield response.follow(
+                rel_url,
+                meta={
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/su.py
```diff
@@ -0,0 +1,63 @@
...
+            )
+
+    def extract_links(self, response):
+        a_tags = response.css("#course-plan-container a")
+        for a_tag in a_tags:
+            url = a_tag.css("::attr(href)").get()
+            anchor = " ".join([
+                response.meta['source_anchor'],
+                response.css("h1.course-title::text").get(),
+                *a_tag.css(" ::text").getall()
```
HarrisonGregg: Is seems we've established that the leading space doesn't do anything here.

osp_scraper/spiders/su.py
```diff
@@ -0,0 +1,63 @@
...
+                    'hops_from_seed': response.meta['hops_from_seed'] + 1,
+                    'source_url': response.url,
+                    'source_anchor': anchor
+                },
+                callback=self.parse_for_files
+            )
+
+        for page in range(2, jsonresponse['total_pages'] + 1):
+            yield scrapy.FormRequest(
+                self.start_urls[0],
```
HarrisonGregg: Is there a reason you used this instead of the usual `response.url`?

osp_scraper/spiders/su.py
```diff
@@ -0,0 +1,63 @@
...
+                meta={
+                    'depth': response.meta['depth'] + 1,
+                    'hops_from_seed': response.meta['hops_from_seed'] + 1,
+                    'source_url': response.url,
+                    'source_anchor': anchor
+                },
+                callback=self.parse_for_files
+            )
+
+        for page in range(2, jsonresponse['total_pages'] + 1):
```
HarrisonGregg: It seems like you're yielding a request for every page on every page.  Scrapy filters them out as duplicates, but you should probably refactor to use a `parse_for_pages` pattern.

osp_scraper/spiders/su.py
jodizzle: This request is a `GET` request which calls back recursively to `parse`, so I need to use a fixed URL.

osp_scraper/spiders/su.py
jodizzle: Woah, that's pretty bad of me.  Thanks for catching that.

jodizzle:
@HarrisonGregg I think I addressed your comments.

osp_scraper/spiders/su.py
HarrisonGregg: Ahh, that makes sense.

osp_scraper/spiders/su.py
```diff
@@ -0,0 +1,65 @@
...
+                meta={
+                    'depth': response.meta['depth'] + 1,
+                    'hops_from_seed': response.meta['hops_from_seed'] + 1,
+                    'source_url': response.url,
+                    'source_anchor': anchor
+                },
+                callback=self.parse_for_files
+            )
+
+        if jsonresponse['page'] >= jsonresponse['total_pages']:
```
HarrisonGregg: As has come up before, it seems more robust to me to yield all the requests at the beginning when possible (like you had originally), instead of chaining them synchronously like this.  But it's up to you.

jodizzle:
@HarrisonGregg Ack, sorry, I keep making that asynchronous vs. synchronous mistake.  Does what I just pushed look more appropriate?

PR #119, opened 2017-06-11
Simplify requirements.txt

jodizzle:
There's no need to have requirements.txt explicitly detail every
dependency -- we can let `pip` take care of most of that.  Additionally,
remove strict version specifications (`==`) for everything but Scrapy,
since most of the time we benefit from having the most recent packages.
The packages `ipython` and `rq-dashboard` aren't strictly necessary, but
are helpful.

Made with the help of [`pigar`](https://github.com/damnever/pigar).


wearpants:
Actually, best practice says we should be pinning exact versions here -
this is an application not a library. Happy to discuss on call

On Jun 11, 2017 2:52 PM, "jodizzle" <notifications@github.com> wrote:

@jodizzle <https://github.com/jodizzle> requested your review on:
opensyllabus/osp-scraper#119
<https://github.com/opensyllabus/osp-scraper/pull/119> Simplify
requirements.txt.

—
You are receiving this because you were mentioned.

<https://github.com/opensyllabus/osp-scraper/pull/119#event-1118612863>,
or mute
the thread

jodizzle:
Hmm, I hadn't considered that, but I think that makes sense.  I can set strict versions.

But in that case, does it make sense to undo the majority of changes in this PR and be more explicit about all dependencies?  That seems to be what's suggested in some places, like here: https://packaging.python.org/requirements/#requirements-files.

HarrisonGregg:
The way I've always seen it done is to just `pip freeze > requirements.txt`.  That makes sure that everyone is running the exact same version of everything and we don't get any inconsistent behavior because of small changes.  When we want to take advantage of the most recent features, we can easily upgrade explicitly.  What advantage is there to not specifying versions?

jodizzle:
@HarrisonGregg I'm familiar with the `pip freeze` approach.  My original intent was to make the requirements file less restrictive so that it would be easier to get the latest updates and we wouldn't have to manually update it as much, but now I realize that doesn't make much sense, especially if the goal is a reproducible virtual environment.

My second statement was about whether we should specify all the implicit dependencies of e.g., Scrapy, or let `pip` handle that itself (for Scrapy, it would basically collect everything like [this](https://github.com/scrapy/scrapy/blob/master/requirements-py3.txt)).

In any case, `requirements.txt` could use some tuning (`celery==3.1.23` is no longer required, for instance), so I could at least repurpose this PR to do that.

jodizzle:
@wearpants @HarrisonGregg Alright, I went ahead and specified all versions of all packages.  Using `pip freeze --local` inside the virtual environment created with the `requirements.txt` file at the previous commit helped.

This should give us the latest versions of everything.  As far as I can tell, scripts and scrapers still run, though it would be nice if someone else could test.

HarrisonGregg:
I'll install these versions and work on scrapers for a while to see if I encounter any problems.

jodizzle:
Seems good then.  Merging.

PR #120, opened 2017-06-12
Add International Christian University scraper

jodizzle:
Similar to #116.

### DIFF
--- a/osp_scraper/spiders/icu.py
+++ b/osp_scraper/spiders/icu.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders.CustomSpider import CustomSpider
+
+class ICUSpider(CustomSpider):
+    name = "icu"
+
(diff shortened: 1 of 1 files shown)


osp_scraper/spiders/icu.py
```diff
@@ -0,0 +1,74 @@
...
+                    'year': year
+                },
+                callback=self.parse_for_pages
+            )
+
+    def parse_for_pages(self, response):
+        for item in self.parse_for_files(response):
+            yield item
+
+        next_page_num = str(response.meta['hops_from_seed'] + 1)
```
HarrisonGregg: It bothers me a little bit that you're relying on `hops_from_seed` here for this.  Probably not going to happen, but theoretically, functionality could be added somewhere that changes `hops_from_seed`.  I'd just use a separate `meta['page']`.

jodizzle: That's a good point.  I'll make the change.

PR #122, opened 2017-06-18
Add Karel De Grote Hogeschool scraper

jodizzle:
This site was pretty uncooperative.


osp_scraper/spiders/kdg.py
```diff
@@ -0,0 +1,112 @@
...
+    def parse(self, response):
+        viewstate = response.css("#__VIEWSTATE::attr(value)").get()
+        viewstategenerator = response.css("#__VIEWSTATEGENERATOR::attr(value)").get()
+        eventvalidation = response.css("#__EVENTVALIDATION::attr(value)").get()
+
+        # We could search here by year and department, but the year selection
+        # can be used to refresh the department list, so we do that first.
+        years = response.css("#MainContent_ddlAcaj option::attr(value)").getall()
+        for year in years:
+            # NOTE: Couldn't get from_response to work here.
```
HarrisonGregg: I may be missing something, but this seems to work for me:
```python
            yield scrapy.FormRequest.from_response(
                response,
                method="POST",
                formdata={
                    '__EVENTTARGET': "ctl00$MainContent$ddlAcaj",
                    'ctl00$MainContent$ddlAcaj': year,
                    'ctl00$btnHome': None
                },
                meta={
                    'year': year
                },
                callback=self.parse_for_departments
            )
```

jodizzle: Yep, that seems to work.  Must've missed it during testing.

jodizzle:
@HarrisonGregg

osp_scraper/spiders/kdg.py
```diff
@@ -0,0 +1,112 @@
...
+    def parse_for_programs(self, response):
+        viewstate = response.css("#__VIEWSTATE::attr(value)").get()
+        viewstategenerator = response.css("#__VIEWSTATEGENERATOR::attr(value)").get()
+        eventvalidation = response.css("#__EVENTVALIDATION::attr(value)").get()
+
+        a_tags = response.css("a[id^='MainContent_lvwProgram_OlodSelect_']")
+        for a_tag in a_tags:
+            # hrefs look like this:
+            #   javascript:__doPostBack('<eventtarget>','')
+            eventtarget = a_tag.css("::attr(href)").re_first(".*\('(.*)',.*")
```
HarrisonGregg: Including `.*` at the beginning and end of the `re_first` regex is unnecessary a makes it a little harder to read.  You could even get away with just `re_first("'(.*)'")` if you wanted.

osp_scraper/spiders/kdg.py
```diff
@@ -0,0 +1,112 @@
...
+            a_text = a_tag.css("::text").get()
+            anchor = response.meta['anchor'] + " " + a_text
+
+            # NOTE: Couldn't get from_response to work here.
```
HarrisonGregg: This seems to work for me:
```python
            yield scrapy.FormRequest.from_response(
                response,
                method="POST",
                formdata={
                    '__EVENTTARGET': eventtarget,
                    'ctl00%24btnHome': None
                },
                meta={
                    'depth': 3,
                    'hops_from_seed': 3,
                    'source_url': response.url,
                    'source_anchor': anchor
                },
                callback=self.parse_for_files
            )
```

jodizzle: Weird, could've sworn I tried this combination of `formdata`, but maybe I was doing it wrong.  Using `ctl00$btnHome` instead of `ctl00%24btnHome` should work just as well, right?

osp_scraper/spiders/kdg.py
jodizzle: Using `re_first("'(.*)'")` leaves behind some other pieces of the original string, like single quotes and a comma, but the slight modification `re_first("'(.*)',")` should eliminate that issue.

jodizzle:
@HarrisonGregg

osp_scraper/spiders/kdg.py
HarrisonGregg: Ahh. Good catch.  Regex was being too _greedy_.  I think you could also fix it by switching to `.*?`.

osp_scraper/spiders/kdg.py
HarrisonGregg: Oh interesting.  Actually, I'm surprised that leaving it as `%24` did work.  Should be `$` instead.

osp_scraper/spiders/kdg.py
jodizzle: Hmm, I suppose that's a bit cleaner.

PR #123, opened 2017-06-19
Remove unecessary scripts

jodizzle:
Branched off from #79.

Deletes `smoke_test_crawler.py` and `csv_prefix_crawler.py` (and the associated method used only in that script, `url_to_prefix_params`).  Also provides minimal documentation updates.


wearpants:
great

PR #124, opened 2017-06-20
Improve parameter creation

jodizzle:
Branched off from #79.

Currently, parameters for crawls done by `FilterSpider` are created through invoking the `osp_scraper.tasks.make_params` method inside `edu_repo_crawler.py`.  Three parameters are created and sent to `FilterSpider` as keyword arguments: `start_urls`, `allowed_file_types` and `filters`.  However, creating the `filters` parameter is the only part of this process that is novel: `start_urls` come directly from the spreadsheet and `allowed_file_types` come from a static class variable.  Given this, it seems reasonable to turn `make_params` into `make_filters`, and further, have `FilterSpider` create it's own parameters with a `from_crawler` class method.  This has the additional benefit of making `FilterSpider` easier to subclass.


bin/edu_repo_crawler.py
```diff
@@ -42,8 +43,10 @@ def main(csv_file, local, institution):
 
                 if urls:
                     log.info("Found %d URLs for %s", len(urls), row['name'])
-                    params = make_params(urls)
-                    log.debug("Parameters: %r", params)
+                    params = {
+                        'start_urls': urls,
```
HarrisonGregg: Minor: trailing comma

bin/edu_repo_crawler.py
```diff
@@ -42,8 +43,10 @@ def main(csv_file, local, institution):
...
+                    }
+                    log.debug("Params: %r", params)
```
HarrisonGregg: Was there a particular reason for changing this line?

osp_scraper/filterware.py
```diff
@@ -1,16 +1,85 @@
...
+            hostname=blacklist_re
+        )
+    )
+
+    # merge parameters from several seed urls, with unique domains & paths
+    for url in seed_urls:
+        u = urllib.parse.urlparse(url)
+        prefix = re.escape(u.path if u.path.endswith('/') else os.path.dirname(u.path) + '/')
+        hostname=re.escape(u.hostname) if u.hostname is not None else None
+        port=re.escape(str(u.port)) if u.port is not None else None
```
HarrisonGregg: Might as well fix the spacing here.

osp_scraper/filterware.py
```diff
@@ -1,16 +1,85 @@
...
+# a list of base domains to blacklist; subdomains blocked too
+blacklist_domains = [
+    'facebook.com',
+    'reddit.com',
+    'twitter.com',
+    'linkedin.com',
+    'wikipedia.org',
+]
+
+blacklist_re = "^" + "|".join(
```
HarrisonGregg: I think you could refactor this while you're moving it.  It creates unnecessarily complex regex.  You could simplify it to something like:
```python
blacklist_re = r"^.*\.({})$".format("|".join(map(re.escape, blacklist_domains)))
```
You could also break it up into more lines for readability if you wanted.

osp_scraper/filterware.py
```diff
@@ -1,16 +1,85 @@
...
+
+    # merge parameters from several seed urls, with unique domains & paths
+    for url in seed_urls:
+        u = urllib.parse.urlparse(url)
+        prefix = re.escape(u.path if u.path.endswith('/') else os.path.dirname(u.path) + '/')
+        hostname=re.escape(u.hostname) if u.hostname is not None else None
+        port=re.escape(str(u.port)) if u.port is not None else None
+
+        # allow prefix to infinite depth
+        filters.append(Filter.compile('allow',
```
HarrisonGregg: The spacing for all `append`s should be consistent.

osp_scraper/filterware.py
```diff
@@ -1,16 +1,85 @@
...
+            pattern='regex',
+            hostname=blacklist_re
+        )
+    )
+
+    # merge parameters from several seed urls, with unique domains & paths
+    for url in seed_urls:
+        u = urllib.parse.urlparse(url)
+        prefix = re.escape(u.path if u.path.endswith('/') else os.path.dirname(u.path) + '/')
+        hostname=re.escape(u.hostname) if u.hostname is not None else None
```
HarrisonGregg: If `u.hostname` is `None`, doesn't that mean that the filters would allow any hostname?  Maybe there's a check somewhere else that prevents it from being a problem, but it seems like this could lead to unbound crawls with the wrong input.

bin/edu_repo_crawler.py
jodizzle: Don't think so.

osp_scraper/filterware.py
jodizzle: Hmm, I think we need to look more into the code that handles the filters to determine if it's really an issue, but I think you're right that in principle it's bad and it would indicate a problem input, similar to how we now raise a `ValueError` for an empty `seed_urls`.  Should we do some more error-raising here?

osp_scraper/filterware.py
jodizzle: Thanks, your regular expression looks nice.  Should `blacklist_re` be built to match leading characters like `www.` though?  The regular expression gets stored in the corresponding filter's `hostname` slot, which I think is supposed to correspond `hostname` a la `urllib.parse.urlparse(url).hostname`, which strips away the `www.`.

Also, is there any reason for `blacklist_re` to be global variable?

HarrisonGregg: Good call.  It looks like `hostname` includes the subdomains, but if there is no subdomain in the first place, we still need to match it.  So it would need to be something like:
```python
blacklist_re = r"(^.*\.)?({})$".format("|".join(map(re.escape, blacklist_domains)))
```
I don't see any reason for it to be global, especially if it's only used once.

osp_scraper/filterware.py
HarrisonGregg: I think that's probably a good idea.

osp_scraper/filterware.py
jodizzle: This seems like the right approach.

wearpants:
looks broadly ok :-)

jodizzle:
@HarrisonGregg I was considering your [comments about checking `u.hostname`](https://github.com/opensyllabus/osp-scraper/pull/124#discussion_r122874053), and I'm wondering a bit more broadly about what types of inputs we'd want to send into `make_filters`.  For example, `u.hostname` will return `None` for relative URLs -- would we ever want to send `make_filters` relative URLs?  Or run `urljoin` on URLs from `seed_urls`?  The list of fields we could consider incorporating into filters or raising errors based on are available in a table in [the `urlparse` documentation](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse).

I'm also wondering about the design of Filters in general; should `None` really mean that "that field is ignored", as it says in the Filter class documentation?  Admittedly, some of these may be questions for other PRs.

PR #125, opened 2017-06-22
Add Institute of Technology Tralee scraper

jodizzle:
Straightforward.


osp_scraper/spiders/ittralee.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
+
+from urllib.parse import urljoin
```
HarrisonGregg: Didn't we decide on not using `from ... import`?

osp_scraper/spiders/ittralee.py
```diff
@@ -0,0 +1,54 @@
...
+        "http://www.ittralee.ie/en/ModDetails/PHPfunction/getsearchfields.php?searchMode=subject"
+    ]
+
+    def parse(self, response):
+        base_url = "http://www.ittralee.ie/en/ModDetails/PHPfunction/module-search-results.php"
+
+        subjs = response.css("#subj_code option")
+        for subj in subjs:
+            subj_value = subj.css("::attr(value)").get()
+            anchor = subj.css("::text").get()
```
HarrisonGregg: Unused?

osp_scraper/spiders/ittralee.py
```diff
@@ -0,0 +1,54 @@
...
+
+        rows = response.css("table.select-module tr")
+        for row in rows:
+            a_tag = row.css("a")
+            if a_tag:
+                rel_url = a_tag.css("::attr(href)").get()
+                anchor = " ".join(row.css("td ::text").getall())
+
+                yield scrapy.Request(
+                    urljoin(base_url, rel_url),
```
HarrisonGregg: It's less explicit, but you could remove a few lines of code and an import with a trick like this:
```python
                yield response.follow(
                    "../" + rel_url,
```

osp_scraper/spiders/ittralee.py
jodizzle: My mistake, this was from when I thought I had to get the subject codes from search form.

osp_scraper/spiders/ittralee.py
jodizzle: No, I think I prefer the more explicit approach.  Even though it's snazzy to use `response.follow`, this seems more confusing to me.

osp_scraper/spiders/ittralee.py
jodizzle: In a recently merged scraper (#115), I used `from urllib.parse import urljoin`, and so for consistency I think I'd like to stick with it for this scraper.  But you're right that we need to consolidate some style conventions, perhaps in another PR soon.

jodizzle:
@HarrisonGregg I think I addressed your comments.

PR #136, opened 2017-07-21
Integrate check for http(s) scheme into filters

jodizzle:
Since 9b2fa8bc985d5b68f66eb8325b2fab370c4d145d, only http(s) links are
permitted during general crawls.  This is good, but the scheme check is
done inside `FilterSpider`, whereas it could be naturally handled by the
filtering system alongside other checks against requests.  Remove the
conditonal from `FilterSpider` and add a scheme filter to
`make_filters`.

Based on the approach described by @HarrisonGregg [here](https://github.com/opensyllabus/osp-scraper/pull/124#issuecomment-312498843).


wearpants:
This was done deliberately, both for efficiency purposes and because the `bogus://` urls cause unnecessary errors in the pipelines. It's really no different than only selecting `href`s instead of all linked objects.

Only thing this would let us do is add new schemes via filters & I can't see doing that often enough (or ever, really).

PR #137, opened 2017-07-25
Change 'osp-site-scrapers' dependency to use HTTPS

jodizzle:
Using HTTPS is slightly easier for people who don't have an ssh-agent or
something like that set up.


HarrisonGregg:
Can you point me to any documentation about how these work?  I don't know enough about pip requirements to understand the difference.

jodizzle:
@HarrisonGregg Here's some general documentation about `pip` with version control: https://pip.pypa.io/en/latest/reference/pip_install/#vcs-support

The basic idea is that, similar to when you clone a private repository from GitHub, you have two choices for authentication: SSH and HTTPS.  Using SSH requires an SSH key registered with your GitHub account, and an SSH agent running on the computer you're trying to clone with.  Using HTTPS requires that your GitHub username and password be input at the commandline.

When [osp-deploy](https://github.com/opensyllabus/osp-deploy) installs osp-scraper, it'll have to use one of these two protocols to pull the dependency.  Originally, SSH was used, the idea being that we could set up SSH keys for deployment.  The idea with this PR is that HTTPS, even if it is a bit more annoying for deployment (deployment becomes less automatic because you have to enter a username/pass), is nicer for developers of osp-scraper (they don't need SSH keys).  It's just a minor change.

HarrisonGregg:
Hmm.  Okay.  In that case, I don't know enough about osp-deploy to understand which one would be easier to use for deployment.  If it's easier to use ssh, then we could just leave it as ssh and mention in the README that you can install it with https first before installing the requirements.

Also, it seems strange to have a public repository that requires access to a private repository to install properly, even if we don't expect anyone else to try using it for anything.  Is there a way we could make it an optional dependency?  Ideally, we'd have no mention of the private repo in the public repo, and we could just have an extra step to install the extra spiders, which would be accessed by the names specified in the spreadsheet, but we could also just have `osp_scraper.spiders` try to import the spiders from `osp_site_scrapers` and fail silently if it doesn't.

HarrisonGregg:
@jodizzle Do you want to use this PR to add the documentation on how to install `osp_site_scraper` with HTTPS when installing for development?

jodizzle:
@HarrisonGregg I think I'd rather just do documentation stuff in a separate PR at some point.  Does that sound alright?

HarrisonGregg:
@jodizzle Sounds good, I just didn't want to close it if you were still planning on using it

PR #139, opened 2017-07-27
Tweaks and improvements for `csv_diff.py`

jodizzle:
- Always preserve `database_urls` if there's a custom scraper.
- Adds a `--diff_custom_scrapers` option, which considers custom scrapers as part of the diff (i.e., custom scrapers that are in both OLD_CSV and NEW_CSV won't be run again).
- Compare rows based on 'id', not 'name'.
- Documentation updates.

If there are any other ideas for tweaking or improving the script, this would be a good place to implement them.


jodizzle:
I can't think of any other small changes that we might want to make in this PR, so I'm gonna go ahead and merge this.

PR #140, opened 2017-08-13
Fix spider middleware yielding requests twice

HarrisonGregg:
This bug currently causes a duplicate request warning message whenever a request that doesn't include `depth` in `request.meta` is also allowed by the filters.  When `request.dont_filter=True` is used, the bug causes `KeyError`s that prevent the spider from functioning.  This change fixes both problems.

### DIFF
--- a/osp_scraper/spidermiddlewares.py
+++ b/osp_scraper/spidermiddlewares.py
@@ -28,6 +28,7 @@ class FilterMiddleware(object):
                 # some requests for files come from scrapy itself (robots.txt, etc.)
                 if 'depth' not in request.meta or request.dont_filter:
                     yield request
+                    continue
 
                 allowed, filter = check_filters(spider.filters, request)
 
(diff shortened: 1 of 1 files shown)


HarrisonGregg:
@wearpants Any comments before I merge this?

wearpants:
Will look today or tomorrow

wearpants:
This is a really good catch! I'd prefer the fix be written as an nested if-else than a continue though, i think that'll make it clearer what's going on.

HarrisonGregg:
For the record, @jodizzle found it.  I had looked for the cause of these problems before, but couldn't find it.

I wasn't sure whether to go for minimal changes, or to refactor for clarity while I was in there.  I've refactored a little and added comments to make it more explicit what's going on.

HarrisonGregg:
@wearpants How does it look now?

wearpants:
This is awesome. :-) :-) :-)

Much clearer than the original version - the bug just drops right out. Nice.

PR #141, opened 2017-08-15
Add scraper for making POST requests

HarrisonGregg:
This scraper takes URLs in the `database_urls` column, then makes a POST request to them, with everything after `?POST_BODY=` as the body of the request.  I've included a csv file for easy testing:
```
bin/edu_repo_crawler.py post_test.csv --local
```
Currently, it just tests [syllabus.kyushu-u.ac.jp](syllabus.kyushu-u.ac.jp) but I'll add more as I find them.

### DIFF
--- a/osp_scraper/spiders/post.py
+++ b/osp_scraper/spiders/post.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+
+import scrapy
+
+from ..spiders import FilterSpider
+
+class PostSpider(FilterSpider):
+    """
+    This spider can submit the given URLs as x-www-form-urlencoded POST
+    requests, and then continues as a general crawl. This avoids having to
+    create custom mapping spiders for many different sites.
+    """
+    name = "post"
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        kwargs['start_urls'] = kwargs['database_urls']
+        spider = super().from_crawler(crawler, *args, **kwargs)
+        return spider
+
+    def start_requests(self):
+        for start_url in self.database_urls:
+            url, body = start_url.split("?POST_BODY=", 1)
+            yield scrapy.FormRequest(
+                url,
+                method="POST",
+                headers={
+                    'Content-Type': "application/x-www-form-urlencoded"
+                },
+                body=body,
+                meta={
+                    'source_url': url,
+                    'source_anchor': body
+                },
+                callback=self.parse
+            )
--- a/osp_scraper/spiders/post.py
+++ b/osp_scraper/spiders/post.py
@@ -12,12 +12,6 @@ class PostSpider(FilterSpider):
     """
     name = "post"
 
-    @classmethod
-    def from_crawler(cls, crawler, *args, **kwargs):
-        kwargs['start_urls'] = kwargs['database_urls']
-        spider = super().from_crawler(crawler, *args, **kwargs)
-        return spider
-
     def start_requests(self):
         for start_url in self.database_urls:
             url, body = start_url.split("?POST_BODY=", 1)
--- a/osp_scraper/spiders/__init__.py
+++ b/osp_scraper/spiders/__init__.py
@@ -73,7 +73,9 @@ class FilterSpider(OSPSpider):
         spider = super().from_crawler(crawler, *args, **kwargs)
         spider.allowed_file_types = ALLOWED_FILE_TYPES
         spider.filters = make_filters(
-            getattr(cls, 'start_urls', []) + kwargs.get('start_urls', []),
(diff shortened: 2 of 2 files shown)


osp_scraper/spiders/post.py
```diff
@@ -0,0 +1,38 @@
...
+                headers={
+                    'Content-Type': "application/x-www-form-urlencoded"
+                },
+                body=body,
+                meta={
+                    'depth': 0,
+                    'hops_from_seed': 0,
+                    'source_url': start_url,
+                    'source_anchor': start_url
+                }
```
jodizzle: Should we be including `callback=self.parse` or `callback=super().parse` for these mappers to make it more clear what's going on?

bin/edu_repo_crawler.py
```diff
@@ -22,7 +22,7 @@ def main(csv_file, local, institution):
         for row in csv.DictReader(f):
             if not institution or institution == row['id']:
                 # Run custom scraper, but only if not running locally.
-                if not local and row['Custom Scraper Name']:
+                if row['Custom Scraper Name']:
```
HarrisonGregg: I should have mentioned, I didn't mean for this to be a change to actually make in this PR, I just included it so we could test the new spider easily.  We could refactor the script to work for both use cases, but I didn't feel like doing it right now.

osp_scraper/spiders/post.py
```diff
@@ -0,0 +1,38 @@
...
+    This spider can submit the given URLs as x-www-form-urlencoded POST
+    requests, and then continues as a general crawl. This avoids having to
+    create custom mapping spiders for many different sites.
+    """
+    name = "post"
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        kwargs['start_urls'] = kwargs['database_urls']
+        spider = super().from_crawler(crawler, *args, **kwargs)
```
HarrisonGregg: That's a good point.  The only reason we didn't do it before is because we'd never thought about a spider that was both a mapping spider and a spider that used `database_urls`, but I don't see any reason not to change that now.

osp_scraper/spiders/post.py
```diff
@@ -0,0 +1,38 @@
...
+                method="POST",
+                headers={
+                    'Content-Type': "application/x-www-form-urlencoded"
+                },
+                body=body,
+                meta={
+                    'depth': 0,
+                    'hops_from_seed': 0,
+                    'source_url': start_url,
+                    'source_anchor': start_url
```
HarrisonGregg: `start_url` does include the body.

osp_scraper/spiders/post.py
HarrisonGregg: I guess including `callback=self.parse` is reasonable.

osp_scraper/spiders/post.py
jodizzle: Wups, my bad.

jodizzle: In that case, should `source_url` be set to `start_url`?  Or should it be the `url` that came from the result of the split?

bin/edu_repo_crawler.py
jodizzle: Sounds fine.

osp_scraper/spiders/post.py
HarrisonGregg: It wouldn't really be the source URL either way, but I guess it would make sense to put `source_url` as `url` and `source_anchor` as `body`.

osp_scraper/spiders/post.py
```diff
@@ -0,0 +1,32 @@
...
+            yield scrapy.FormRequest(
+                url,
+                method="POST",
+                headers={
+                    'Content-Type': "application/x-www-form-urlencoded"
+                },
+                body=body,
+                meta={
+                    'depth': 0,
+                    'hops_from_seed': 0,
```
jodizzle: Should these values be `1` instead of `0`?  I always get a bit confused by this...

HarrisonGregg: I don't think so.  `start_requests` in `FilterSpider` sets them to `0`.

HarrisonGregg:
I'm going to wait until I try this with some other sites before I merge.

jodizzle:
I rebased to remove the test file, remove the `depth`/`hops_from_seed` counters (per https://github.com/opensyllabus/osp-scraper/pull/160), and otherwise updated with regards to `master`.

After testing on a few sites, I can confirm that this feature works, and has the potential to be very useful.  Any other concerns about it, or developing similar solutions for quicker grabbing of sites (e.g., could we use an approach like this for sites that require a single GET form request?), can be addressed in other tickets.

PR #144, opened 2017-08-18
FilterSpider: Remove `process_file_url` and `process_text_url`

jodizzle:
These methods are identical, so there isn't a need for two of them.
Moreover, they're tiny and single-purpose, so it's much more explicit to
just do the 'meta' creation inline.


wearpants:
I think this looks ok - were those methods overriden anywhere in the codebase?

wearpants:
looks good.

Issue #72, opened 2017-04-17
Scrapy: Issues with Scrapy shell


jodizzle:
When running `scrapy shell` from the dotenv, the first `fetch` (e.g, `fetch('http://example.com/')`) always leads to two successive `AttributeError: 'DefaultSpider' object has no attribute 'run_id'` errors.  This does not happen on further `fetch`es from within the same shell instance.

When using inspect response from a spider script, i.e,
```python
from scrapy.shell import inspect_response
inspect_response(response, self)
```
the shell loads the response as expected, but after exiting the shell, SIGINT (ctrl-C) from the terminal no longer seems to be received by Scrapy as a shutdown signal.

Issue #69, opened 2017-04-17
Scrapy: Priority and the request queue


jodizzle:
As mentioned by @HarrisonGregg [here](https://github.com/opensyllabus/osp-scraper/pull/63#issuecomment-294331359), Scrapy seems to have some strange approaches to prioritizing requests.  Attempts to force priority via the `priority` parameter for `scrapy.Request` doesn't seem to work as intended either.

Could this be because of a some setting we have/don't have?  Could this be related to something that's been addressed in Scrapy, and we need to update the version in `requirements.txt`?

PR #145, opened 2017-08-19
Remove unnecessary spider middleware

HarrisonGregg:
It seems as though `spidermiddleware.py` is just duplicating the functionality of `downloadermiddleware.py`.  Some simple tests without `spidermiddleware.py` indicate that the behavior is unchanged without it.

The only functionality in `spidermiddleware.py` that is not in `downloadermiddleware.py` is resetting `request.depth` to `0`, but I suspect that this was meant to instead reset `request.meta['depth']` to `0`, which is done in `downloadermiddleware.py`.

Either middleware could be used to accomplish the filtering, but it appears that `downloadermiddleware.py` is able to do it much more simply.


wearpants:
Will look next week  but we needed both middlewares because scrapy has 2
independent pipelines (spider and downloader) with somewhat different APIs

Is this causing any problems or bugs as is?

HarrisonGregg:
@wearpants I think you might be confusing some terminology here, which leads to some confusion about how these different Scrapy components interact.  After reading through the Scrapy documentation, this is how I understand it.

Scrapy allows you to use [Item Pipelines](https://doc.scrapy.org/en/latest/topics/item-pipeline.html) such as `FilesPipeline`, `ImagesPipeline`, or, in our case, `WarcFilesPipeline` and `WarcStorePipeline`.  These pipelines process every item that is yielded by a spider and (among other things) generate additional requests for files.  These requests may have different properties (such as priority and callback) from requests generated in spiders and so may behave differently.

Scrapy also allows you to use two different kinds of middleware, [Downloader Middleware](https://doc.scrapy.org/en/latest/topics/downloader-middleware.html) and [Spider Middleware](https://doc.scrapy.org/en/latest/topics/spider-middleware.html).  They offer similar functionality to each other, but have a few differences.  All requests, whether generated by a spider or by the pipeline, go through both Downloader Middleware and Spider Middleware.

`spidermiddleware.py` doesn't appear to be causing any problems.  We're just trying to clean up the codebase as we track down some remaining bugs we've been experiencing (such as #69 and the second half of #72), since we don't know where the bugs might originate.

wearpants:
Yeah sorry it's been over a year since I looked at this scrapy stuff in
depth, I'll review this week

wearpants:
It has been a very long time since I dug into this, but I am 100% sure both of these are necessary - I wrote versions with only one of them and it did not function correctly. The middlewares are invoked at different spots in the cycle & you need both to capture all requests scrapy will generate (redirects, etc.): https://doc.scrapy.org/en/latest/topics/architecture.html Is there a specific problem you've identified that this is causing?

That said, that `request.depth = 0` does look a little dodgy.

jodizzle:
@wearpants It's absolutely true that the different types of middlewares are invoked at different spots in the Scrapy cycle, and that we need to be mindful about all the different types of requests being generated.  However, what is currently in place are two different types of middlewares -- a Spider middleware and a Downloader middleware -- that both check requests against our filtering system.  This seems unnecessary; If you look at the "Data Flow" diagram on the page you linked, all requests, whether generated from the Item pipeline, from our spiders, or from Scrapy, have to eventually pass through the Downloader middleware.  Why can't we just do all the filtering there?  I'm still investigating certain details, and it may be helpful to have some sort of custom Spider middleware in place (to handle some depth calculations, for instance), but I currently don't see any point to having two middlewares in two different places in the cycle that do the exact same thing.

There's no specific problem _per se_ with having both middlewares in place, but errors and oversights like the `request.depth = 0` issue and the multiple yield bug in https://github.com/opensyllabus/osp-scraper/pull/140 indicate that some bugs have already snuck in.  Simplifying the code and making the most out of what Scrapy provides are both important goals and this PR is a good step.

wearpants:
I spent a few weeks going down this exact rabbit hole when I wrote the
thing :-\

There *are* two APIs in scrapy here, and it was frustrating to deal with
both of them... I was dissatisfied with the similar code in 2 middlewares,
and wanted to write a single behavior class + 2 adapter classes, but ran
out of time... I'd be happy with an approach along those lines

On Aug 23, 2017 5:15 AM, "jodizzle" <notifications@github.com> wrote:

@wearpants <https://github.com/wearpants> It's absolutely true that the
different types of middlewares are invoked at different spots in the Scrapy
cycle, and that we need to be mindful about all the different types of
requests being generated. However, what is currently in place are two
different types of middlewares -- a Spider middleware and a Downloader
middleware -- that both check requests against our filtering system. This
seems unnecessary; If you look at the "Data Flow" diagram on the page you
linked, all requests, whether generated from the Item pipeline, from our
spiders, or from Scrapy, have to eventually pass through the Downloader
middleware. Why can't we just do all the filtering there? I'm still
investigating certain details, and it may be helpful to have some sort of
custom Spider middleware in place (to handle some depth calculations, for
instance), but I currently don't see any point to having two middlewares in
two different places in the cycle that do the exact same thing.

There's no specific problem *per se* with having both middlewares in place,
but errors and oversights like the request.depth = 0 issue and the multiple
yield bug in #140 <https://github.com/opensyllabus/osp-scraper/pull/140>
indicate that some bugs have already snuck in. Simplifying the code and
making the most out of what Scrapy provides are both important goals and
this PR is a good step.

—
You are receiving this because you were mentioned.

<https://github.com/opensyllabus/osp-scraper/pull/145#issuecomment-324271146>,
or mute the thread

HarrisonGregg:
@wearpants Can you elaborate on what you mean by two APIs?  I've been trying to better understand how scrapy works, but I don't follow what you mean.

I've been looking into the practical implications of this question further though.  In #158, I've created a test server that includes a number of different cases regarding depth and redirects and different kinds of links, and a test suite that runs against this server.  Using this test server, I've confirmed a few different things:

- The results of the test suite are unaffected by removing the spider middleware
- The contents and metadata of the WARC files generated with and without the spider middleware are functionally identical
- All INFO and DEBUG messages generated while running the tests are identical
- If you have the spider middleware allow all requests, every single request that goes through the spider middleware when crawling the test server also goes through the downloader middleware

Do you remember what sort of issues you were running into that required the spider middleware to address?  Or are there any other relevant edge cases that I should include on the test server?

HarrisonGregg:
Also, the tests confirm that the depth is not always reset properly with only the spider middleware (with the `request.depth` bug fixed) and without the downloader middleware.

wearpants:
How does this relate to #160?

jodizzle:
@wearpants This PR is about removing what seems to be an unnecessary, duplicate `FilterMiddleware` function from our `spidermiddlewares` in favor of the more appropriate `FilterMiddleware` in our `downloadermiddlewares`.  PR #160 is about making a `DepthMiddleware` in our `spidermiddlewares` that would automate the incrementing of `depth` and `hops_from_seed` so that we don't have to manually do it ourselves in the various spiders.

I haven't really done review/testing of #160 yet, but it should be compatible with this PR.

jodizzle:
I rebased to resolve the conflicts (i.e., make compatible with https://github.com/opensyllabus/osp-scraper/pull/160).  Merging now.

PR #148, opened 2017-08-20
Move dont_filter check so depth is reset properly

HarrisonGregg:
Currently, using `dont_filter=True` on a request prevents the depth from being reset in the downloader middleware if the matching filter has no `max_depth`.  This is a simple change that fixes the issue, but should have no other effects.


HarrisonGregg:
@wearpants Actually, we encountered this problem with Jody's [Delft University Spider](https://github.com/opensyllabus/osp-site-scrapers/pull/19).  `depth` wasn't getting reset properly because of this problem.

wearpants:
sounds fine

PR #150, opened 2017-08-24
Add Scrapy templates for our spider types

jodizzle:
Not super useful, just something fun I thought of.

To test, run
```
scrapy genspider -t [scraper|mapper] [name] [start_url]
```

Scrapy templates are pretty limited in their capabilities as far as I can tell, so much of the time the output of the `genspider` command will still need tweaking, but in theory these templates will still _greatly_ optimize our workflow ;)


wearpants:
hey cool

jodizzle:
@HarrisonGregg It's mostly just a fun micro-optimization.  And again, Scrapy's template-ing capabilities are unfortunately pretty limited.

PR #142, opened 2017-08-15
Updates for seed URL crawls

davidmcclure:
This PR makes a couple of changes needed to crawl the lists of miscellaneous seed URLs.

- Add a `queue_seed_urls` script, which reads a file of line-delimited seed URLs, groups them by domain, and queues crawl jobs for each.
- Add a `SeedURLList` class, which wraps the interaction with the file of seed URLs.
- In the filters, for an in-prefix URL, set `max_depth=100`, to avoid crawler traps.
- Add a custom ScraperWorker class that fails tasks when a worker is explicitly shut down. (Long story; basically, this means if a queue worker dies, running jobs gets put on the "failed" queue, instead of just getting lost.)
- Minor - group GitHub dependencies together at the bottom of `requirements.txt`.

PR #151, opened 2017-08-25
Add script for queuing seed URLs

HarrisonGregg:
@davidmcclure I've refactored your script for queuing the seed URLs.  The key changes are:

1.  It seemed a little unnecessary to have an entire file for grouping URLs by domain when the functionality of the file was just a few lines.  If you think that grouping by domain is something we might need to use in other places though, we could also move it into a method in `utils.py`.
2. I added a method to `tasks.py` that makes it easy to specify `timeout` when queuing jobs.  I'll add it to `edu_repo_crawler.py` in a separate PR.

### DIFF
--- a/bin/queue_seed_urls.py
+++ b/bin/queue_seed_urls.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+
+from collections import defaultdict
+import logging
+from urllib.parse import urlparse
+
+import click
+
+from osp_scraper.tasks import get_crawl_job
+
+
+log = logging.getLogger('seed_url_crawler')
+
+
+@click.command()
+@click.argument('path', type=click.Path(exists=True))
+def main(path):
+    groups = defaultdict(list)
+    for url in open(path):
+        domain = urlparse(url.strip()).netloc
+        groups[domain].append(url.strip())
+
+    queue_crawl = get_crawl_job(timeout='24h')
+    for domain, urls in groups.items():
+        queue_crawl(
+            'osp_scraper_spider',
+            start_urls=urls,
+            ignore_robots_txt=True
+        )
+
+        log.info(f'{len(urls)} URLs for {domain}')
+
+if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


jodizzle:
Split off from #142.

bin/queue_seed_urls.py
```diff
@@ -0,0 +1,35 @@
...
+
+from osp_scraper.tasks import get_crawl_job
+
+
+log = logging.getLogger('seed_url_crawler')
+
+
+@click.command()
+@click.argument('path', type=click.Path(exists=True))
+def main(path):
```
jodizzle: I think having a docstring similar to the one on [csv_diff](https://github.com/opensyllabus/osp-scraper/blob/3a6a8513a1de5251e4525589562d9d1e6c4f3cd3/bin/csv_diff.py#L19) would be helpful.

bin/queue_seed_urls.py
```diff
@@ -0,0 +1,35 @@
...
+    groups = defaultdict(list)
+    for url in open(path):
+        domain = urlparse(url.strip()).netloc
+        groups[domain].append(url.strip())
+
+    queue_crawl = get_crawl_job(timeout='24h')
```
jodizzle: Instead of hard-coding a value, I think it would be valuable to have a `click.option` for `timeout` with a reasonable default.

bin/queue_seed_urls.py
HarrisonGregg: Sounds good.  I'm going to start trying to adhere to the docstring format of having a short description on the first line, and then a longer description after a newline.

bin/queue_seed_urls.py
HarrisonGregg: Good idea.

bin/queue_seed_urls.py
jodizzle: Ah, yes.  We (I) should be more consistent about doing this.

davidmcclure:
This is fine with me. 👍

Issue #149, opened 2017-08-24
Make `--local` option in crawl scripts work when running multiple crawls


jodizzle:
In theory, the `--local` option for crawl scripts like `edu_repo_crawler` should run crawls sequentially and without the need for a redis server.  However, it seems like after the first crawl completes, twisted crashes with a `twisted.internet.error.ReactorNotRestartable` error right after starting the next crawl.

[This SO post](https://stackoverflow.com/questions/41495052/scrapy-reactor-not-restartable/43661172#43661172) seems relevant.

wearpants:
Yup, we used to support this under celery but it got taken out

PR #153, opened 2017-08-25
Add ScraperWorker class

jodizzle:
Used by osp-deploy as of
opensyllabus/osp-deploy@9b4a1b7b1d30d32aca5467465b2e661b80a738cd.

From the `deploy` branch, and split off from https://github.com/opensyllabus/osp-scraper/pull/142.

### DIFF
--- a/osp_scraper/workers.py
+++ b/osp_scraper/workers.py
@@ -1,0 +1,20 @@
+
+
+from rq import Worker
+
+
+class ScraperWorker(Worker):
+
+    def request_stop(self, *args, **kwargs):
+        """When SIGINT is sent to the worker (eg, if the Supervisor process
+        group is restarted), immediately fail the running job and stop the
+        worker. This avoids a scenario in which the worker gets shut down but
+        not unregistered in Redis, causing it to get "marooned" in the admin.
+        """
+        job = self.get_current_job()
+
+        if job:
+            self.handle_job_failure(job)
+            self.failed_queue.quarantine(job, 'Worker shutdown.')
+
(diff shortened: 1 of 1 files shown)


jodizzle:
@HarrisonGregg As far as I can tell there's no way to set a default worker class for `rqworker`.  We could create a custom config file that specifies it -- we already have the beginnings of that in some sense with [`osp_scraper.settings.rq`](https://github.com/opensyllabus/osp-scraper/blob/3a6a8513a1de5251e4525589562d9d1e6c4f3cd3/osp_scraper/settings/rq.py) -- but using that config would still require a command line argument for people not using osp-deploy.

I can look into the settings stuff some more, but I'm assuming @davidmcclure knows more than I do.

davidmcclure:
Hey @jodizzle and @HarrisonGregg,

Yeah, this is only relevant when a spider is getting run under RQ in production. Basically, RQ doesn't really anticipate that jobs will be as long-running as the crawls are, and, out of the box, doesn't handle the case where a job gets cut off half-way through in a way that makes sense in this context. In fact, in development, it should almost never be necessary to run spiders in the queue, as opposed to just firing off the crawls directly.

jodizzle:
@davidmcclure Yep, that makes sense; doesn't seem like using the `ScraperWorker` is too necessary for development.  We should still probably add a note about it somewhere, though.

The reason we run a redis server for some of our development now is because currently, using the `--local` option is pretty broken, as per https://github.com/opensyllabus/osp-scraper/issues/149.

PR #155, opened 2017-08-31
Add script for filtering CSVs by custom scraper name(s)

jodizzle:
This script can be used with `edu_repo_crawler.py` to do crawling over
specific 'Custom Scraper Name'(s).  This script is probably most useful
when testing framework scrapers that leverage 'Database URLs', but
should also be helpful for any type of targeted or limited crawling that
may happen future.

We could probably roll up a change like this into `csv_diff.py`, but I feel pretty fine keeping a separate script.

### DIFF
--- a/bin/csv_filter.py
+++ b/bin/csv_filter.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python3
+
+import csv
+
+import click
+
+@click.command()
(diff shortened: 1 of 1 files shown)


jodizzle:
@HarrisonGregg Yeah.  Personally, I don't mind pushing little scripts that do things like this, even if they only have limited use.

PR #156, opened 2017-09-05
Add max_hops_from_seed option to filters

HarrisonGregg:


osp_scraper/filterware.py
```diff
@@ -154,6 +156,9 @@ def __call__(self, request):
                 ret = False
             elif self.max_depth is not None and request.meta['depth'] > self.max_depth:
                 ret = False
+            elif self.max_hops_from_seed is not None\
+                    and request.meta['hops_from_seed'] > self.max_hops_from_seed:
```
jodizzle: Two suggestions:
1) Place the newline after a logical operator.
2) Extend by wrapping inside parenthesis instead of using `\`.

HarrisonGregg: Yeah, I guess parentheses make more sense here.

HarrisonGregg:
@jodizzle  Look good?

osp_scraper/filterware.py
```diff
@@ -154,6 +156,9 @@ def __call__(self, request):
                 ret = False
             elif self.max_depth is not None and request.meta['depth'] > self.max_depth:
                 ret = False
+            elif (self.max_hops_from_seed is not None and
+                request.meta['hops_from_seed'] > self.max_hops_from_seed):
```
jodizzle: Actually, is there a reason you chose to align it like this?

HarrisonGregg: I aligned it like that because the one above on line 155 is aligned like that.  Otherwise, I'd probably indent it one more tab.

jodizzle: Oh, I see.  I interpret line 155 as being aligned with the opening delimiter, but you can leave it like this if you want.

HarrisonGregg: Ahh, good point.  I guess that is the standard way of doing it.

PR #157, opened 2017-09-05
Allow for multiple custom scrapers per row

HarrisonGregg:
Occasionally one institution uses multiple distinct databases that require distinct custom scrapers.  This change allow specifying multiple comma-separated scraper names in the "Custom Scraper Name" column.

However, this brings up concerns about having too big of an impact on servers by running multiple crawls on the same domain at once, so this also makes a change that causes the crawl jobs in each row to run in sequence instead of (potentially) in parallel.  Each job will wait for the previous job to finish before starting.  

@wearpants @davidmcclure Does this sound reasonable to you?  Does it make sense to run the crawls sequentially even though it could mean it takes a lot longer to run all of the crawls?  The custom scraper crawls shouldn't take as long as the general crawls though, so it shouldn't make too big of a difference.

### DIFF
--- a/osp_scraper/tasks.py
+++ b/osp_scraper/tasks.py
@@ -28,6 +28,8 @@ def crawl(spider, *args, **kwargs):
     except KeyError as err:
         # Log a warning if the scraper name is invalid instead of
         # causing the job to fail.
+        # NOTE: If there is any other type of error, the job will fail, and all
+        # the jobs that depend on it will fail as well.
         logger.warning(err.args[0])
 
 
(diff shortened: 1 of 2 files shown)


HarrisonGregg:
@jodizzle It seems like that would actually be a big problem.  It's not inconceivable that an invalid scraper name would be in the Custom Scraper Column, especially since we sometimes add the names to the spreadsheet before the PR is merged.  I've made a change to `tasks.py` that prevent an invalid scraper name from causing the job to fail, instead just logging a warning.  However, this setup still could be problematic if there are other errors that could cause job failure.

osp_scraper/tasks.py
```diff
@@ -25,8 +28,13 @@ def crawl(spider, *args, **kwargs):
...
 
     proc = CrawlerProcess(settings)
-    proc.crawl(spider, *args, **kwargs)
-    proc.start()
+    try:
+        proc.crawl(spider, *args, **kwargs)
+        proc.start()
+    except KeyError as err:
+        # Log a warning if the scraper name is invalid instead of
+        # causing the job to fail
```
jodizzle: Minor: Add a period.

jodizzle:
I rebased to update with respect to master and remove two tiny commits (changing the logger name and removing a period).  I also added an extra log message and some inline comments (hopefully not too verbose).

PR #158, opened 2017-09-07
Add a test suite

HarrisonGregg:
This is a test suite that runs `osp_scraper_spider` on a local test server and then checks the contents of the WARC files that are generated.

I don't think it's ready to be merged yet, but I think it's far enough along to start discussing it here.


wearpants:
Not sure if this will hit all paths to `files_url` / `DownloadPipeline`?

VERY EXCITING

tests/server_tests.py
```diff
@@ -0,0 +1,164 @@
...
+        self.assertIn("/path/0/file/fileA.rtf", self.warc_metadata_by_path)
+
+    def test_max_depth(self):
+        self.assertIn("/path/2/", self.warc_metadata_by_path)
+        self.assertNotIn("/path/3/", self.warc_metadata_by_path)
+
+    def test_max_hops_from_seed(self):
+        self.assertIn("/path/0/infinite/1", self.warc_metadata_by_path)
+
+        self.assertIn("/path/0/infinite/10", self.warc_metadata_by_path)
```
jodizzle: For situations like these, wouldn't it be more appropriate to run asserts on all relative URLs up to `max_hops_from_seed`?

tests/server_tests.py
```diff
@@ -0,0 +1,164 @@
...
+        os.environ["LOG_LEVEL"] = "DEBUG"
+
+        if os.path.isdir(FILES_DIR):
+            shutil.rmtree(FILES_DIR)
+
+        crawl(
+            'osp_scraper_spider',
+            start_urls=[
+                f"{HOST}/path/0/",
+                f"{HOST}/start_path/",
```
jodizzle: Is there a particular reason why `/start_path/` still needs to be crawled?  Could the tests that use it below be reconfigured to just be based on `/path/0/`?

tests/server_tests.py
```diff
@@ -0,0 +1,164 @@
...
+            "/path/0/iframe?url=/path/0/embedded",
+            self.warc_metadata_by_path
+        )
+        self.assertIn("/path/0/embedded", self.warc_metadata_by_path)
+
+    def test_off_domain_depth(self):
+        self.assertIn(
+            "http://httpbin.org/html?source_path=0",
+            self.warc_metadata_by_path
+        )
```
jodizzle: For situations like these, where we're trying to check if depth conditions are being met, I feel like we should be checking for presence of the key _and_ value (`assertIn` and `assertEqual`).

tests/server_tests.py
jodizzle: I'll leave this in for now.  It seems like a good idea to test multiple `start_urls` anyway.

PR #159, opened 2017-09-07
Add support for running multiple spiders locally

### LINKED ISSUE #149: Make `--local` option in crawl scripts work when running multiple crawls

HarrisonGregg:
These changes resolve #149.  It requires a bit of a workaround, but it seems like a reasonably elegant solution.

### DIFF
--- a/bin/edu_repo_crawler.py
+++ b/bin/edu_repo_crawler.py
@@ -5,7 +5,7 @@ import logging
 
 import click
 
-from osp_scraper.tasks import crawl, get_crawl_job
+from osp_scraper.tasks import get_crawl_job, LocalQueue
 from osp_scraper.utils import extract_urls
 
 log = logging.getLogger('edu_repo_crawler')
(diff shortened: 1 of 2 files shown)


jodizzle:
Note that `CrawlerProcess` is just an extension of `CrawlerRunner`, as detailed [here](https://doc.scrapy.org/en/latest/topics/api.html#scrapy.crawler.CrawlerProcess), and comes with some nice bonuses "if you aren’t running another Twisted reactor within your application."

HarrisonGregg:
@jodizzle My impression based on [Running multiple spiders in the same process](https://doc.scrapy.org/en/latest/topics/practices.html#running-multiple-spiders-in-the-same-process) was that you need to use Twisted deferreds to get the spiders to run sequentially.  I did try the simpler approaches to running multiple spiders simultaneously, but given that our use case is to run spiders locally for testing purposes, it seems way too messy to me to have multiple spiders running simultaneously.  I'd prefer to have them running sequentially so it's more clear what's going on.

I thought about using a `queue` or a `deque`, but it didn't seem necssary given the simple use case.

jodizzle:
@HarrisonGregg I see, I missed that.  I had originally thought by looking at your code that what you were doing didn't run the scrapers sequentially, but from testing I now see that that's not the case.  I think using `CrawlerRunner` is fine; we miss out on some of the special features from using `CrawlerProcess` (e.g., shutdown signals), but that's not too useful for our purposes.

jodizzle:
> Other than the [two](https://github.com/opensyllabus/osp-scraper/pull/159/files#r138275315) [comments](https://github.com/opensyllabus/osp-scraper/pull/159/files#r138475154) above, I think this is good.

I've addressed them.

PR #160, opened 2017-09-07
Refactor all depth/hops_from_seed logic into spider middleware

HarrisonGregg:
The primary changes in this PR are to consolidate all logic concerning setting and incrementing `depth` and `hops_from_seed` on requests and items into a new spider middleware.  This simplifies logic somewhat in this project, but will make a more significant impact in the custom scrapers, which will no longer have to include any `depth`/`hops_from_seed` logic at all.  Some code is included for backward compatibility with custom scrapers, but it could be removed after all custom scrapers are updated.

Additionally, this PR removes the line that disables [`scrapy.spidermiddlewares.depth.DepthMiddleware`](https://github.com/scrapy/scrapy/blob/master/scrapy/spidermiddlewares/depth.py).  This middleware does not appear to conflict with our own `depth` logic, and it also sets the priority of requests based on their depth, which should make our spiders behave more consistently.

Testing with the suite in #158 as well as manual tests indicate that these changes do not affect the functionality of any spiders.

### DIFF
--- a/osp_scraper/spidermiddlewares.py
+++ b/osp_scraper/spidermiddlewares.py
@@ -26,7 +26,6 @@ class DepthMiddleware(object):
         hops_from_seed = response.meta.get('hops_from_seed', 0)
 
         for obj in result:
-            print("result:", obj)
             if isinstance(obj, PageItem):
                 obj['depth'] = depth
                 obj['hops_from_seed'] = hops_from_seed
--- a/osp_scraper/spidermiddlewares.py
+++ b/osp_scraper/spidermiddlewares.py
@@ -2,10 +2,52 @@ import logging
 
 from scrapy import Request
 
+from .items import PageItem
 from .filterware import check_filters
 
 logger = logging.getLogger(__name__)
 
+class DepthMiddleware(object):
+    """Middleware to set depths and hops_from_seed
+
+    This middleware handles all logic for setting depth and
+    hops_from_seeds on requests and items.
+    """
+
+    def __init__(self, stats):
+        self.stats = stats
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler.stats)
+
+    def process_spider_output(self, response, result, spider):
+        depth = response.meta.get('depth', 0)
+        hops_from_seed = response.meta.get('hops_from_seed', 0)
+
+        for obj in result:
+            print("result:", obj)
+            if isinstance(obj, PageItem):
+                obj['depth'] = depth
+                obj['hops_from_seed'] = hops_from_seed
+
+                for url, meta in obj['file_urls']:
+                    meta['depth'] = depth + 1
+                    meta['hops_from_seed'] = hops_from_seed + 1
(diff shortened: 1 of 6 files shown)


osp_scraper/spidermiddlewares.py
```diff
@@ -3,10 +3,53 @@
...
+            if isinstance(obj, PageItem):
+                obj['depth'] = depth
+                obj['hops_from_seed'] = hops_from_seed
+
+                for url, meta in obj['file_urls']:
+                    meta['depth'] = depth + 1
+                    meta['hops_from_seed'] = hops_from_seed + 1
+
+            elif isinstance(obj, Request):
+                obj.meta['depth'] = depth + 1
```
HarrisonGregg: By the way, I think this is the only line that duplicates functionality in `scrapy.spidermiddlewares.depth.DepthMiddleware`.

osp_scraper/spidermiddlewares.py
```diff
@@ -3,10 +3,53 @@
...
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler.stats)
+
+    def process_spider_output(self, response, result, spider):
+        depth = response.meta.get('depth', 0)
+        hops_from_seed = response.meta.get('hops_from_seed', 0)
+
+        for obj in result:
+            print("result:", obj)
```
wearpants: let's remove this stray `print`

HarrisonGregg: Good catch.

osp_scraper/spidermiddlewares.py
```diff
@@ -2,10 +2,51 @@
...
+
+            yield obj
+
+    def process_spider_input(self, response, spider):
+        # This is only necessary for backwards compatibility in custom
+        # scrapers that expect depth and hops_from_seed in response.meta
+        if 'depth' not in response.meta:
+            response.meta['depth'] = 0
+        if 'hops_from_seed' not in response.meta:
+            response.meta['hops_from_seed'] = 0
```
jodizzle: Are we sure that these clauses are required?  I'm having a hard time finding a case where they're necessary.  What exactly does the comment above
```python
# This is only necessary for backwards compatibility in custom
# scrapers that expect depth and hops_from_seed in response.meta
```
mean?

HarrisonGregg: It won't be necessary once we merge [this PR](https://github.com/opensyllabus/osp-site-scrapers/blob/master/osp_site_scrapers/hik.py#L35), but until then, scrapers like [this one](https://github.com/opensyllabus/osp-site-scrapers/blob/master/osp_site_scrapers/hik.py#L35) will throw an error if `depth` and `hops_from_seed` aren't in the initial requests.  We can leave it out if we're definitenly merging the other PR immediately after this one.

jodizzle: Ah, my bad.  I had been looking at the wrong examples.

PR #161, opened 2017-09-08
Clean up imports and remove unnecessary encoding declaration

HarrisonGregg:
This PR:
- Removes unused imports
- Organize imports per PEP8 guidelines
- Removes `# -*- coding: utf-8 -*-` from a few files, since it's unnecessary in Python 3
- Remove blank lines at the beginning of files
- Fix spacing so that there is 1 line after imports, but 2 lines surrounding top level class and function declarations, per PEP8
- Remove `httpstatus35` requirement and replace usage with `http` from standard library

### DIFF
--- a/osp_scraper/templates/spiders/mapper.tmpl
+++ b/osp_scraper/templates/spiders/mapper.tmpl
@@ -1,8 +1,7 @@
-# -*- coding: utf-8 -*-
-
 import scrapy
 
 from osp_scraper.spiders import FilterSpider
+
 
 class $classname(FilterSpider):
     name = "$name"
--- a/osp_scraper/templates/spiders/scraper.tmpl
+++ b/osp_scraper/templates/spiders/scraper.tmpl
@@ -1,8 +1,7 @@
-# -*- coding: utf-8 -*-
-
 import scrapy
 
 from osp_scraper.spiders.CustomSpider import CustomSpider
+
 
 class $classname(CustomSpider):
     name = "$name"
--- a/bin/csv_diff.py
+++ b/bin/csv_diff.py
@@ -6,6 +6,7 @@ import click
 
 from osp_scraper.utils import extract_urls
 
+
 @click.command()
 @click.argument('old_csv_file', type=click.Path(exists=True))
 @click.argument('new_csv_file', type=click.Path(exists=True))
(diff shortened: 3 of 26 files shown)


osp_scraper/filters.py
```diff
@@ -16,7 +16,6 @@
     'wikipedia.org',
 ]
 
-
```
jodizzle: PEP 8 specifies that we ["surround top-level function and class definitions with two blank lines"](https://www.python.org/dev/peps/pep-0008/#blank-lines), so since this is the top-level function, I think you should leave the second blank line.

osp_scraper/templates/spiders/mapper.tmpl
```diff
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 import scrapy
 
 from osp_scraper.spiders import FilterSpider
```
jodizzle: Add an extra blank line after this (PEP 8).

osp_scraper/templates/spiders/scraper.tmpl
```diff
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 import scrapy
 
 from osp_scraper.spiders.CustomSpider import CustomSpider
```
jodizzle: Add an extra blank line after this (PEP 8).

HarrisonGregg: Good catch.

HarrisonGregg:
@jodizzle I added another commit to address your question [here](https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/settings/scrapy.py#L114).

Alternatively, we could do it the way that scrapy does it [in the default settings](https://github.com/scrapy/scrapy/blob/bbc56e6960ff4b9ea824bf8c85a9e29b1960ba3e/scrapy/settings/default_settings.py#L269):
```python
from os.path import join, abspath, dirname
TEMPLATES_DIR = abspath(join(dirname(__file__), '..', 'templates'))
```
but I think the approach I commited is cleaner.

Is that alright with you?

jodizzle:
@HarrisonGregg That seems fine.  Thanks.

PR #162, opened 2017-09-09
Fix iteritems -> items, for Python 3.

davidmcclure:
Landing a fix that Jody and I made a couple weeks ago on the deploy branch, which I just noticed never made it into master - in the `pipelines` module, there was an `.iteritems()` call on a dictionary, which breaks in Python 3, where it's just `.items()`.

This is the last unmerged change on `deploy`, which I'll get rid of. Going forward let's just deploy from master.

### DIFF
--- a/osp_scraper/pipelines.py
+++ b/osp_scraper/pipelines.py
@@ -76,7 +76,7 @@ def update_warc_response_from_item(record, item):
     # XXX scrapy doesn't provide human-readable status string
     status = "HTTP/1.1 {} {}".format(item['status'],
                                      httpstatus.HTTPStatus(item['status']).name).encode()
-    headers = [b': '.join((k, v)) for k, l in item['headers'].iteritems() for v in l]
+    headers = [b': '.join((k, v)) for k, l in item['headers'].items() for v in l]
 
     record.update_payload(b"\r\n".join(itertools.chain((status, ),
                                                        headers,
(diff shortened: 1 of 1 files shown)


jodizzle:
@davidmcclure This fix is actually unnecessary: the problem was that `item['headers']` is usually of `dict`-like type `scrapy.http.headers.Headers`, whereas in an old version of the `FileDownloader` (what we were invoking when we ran into the bug), [`item['headers']` was set to a regular Python `dict`](https://github.com/opensyllabus/osp-scraper/commit/9e4a55135bb68ef9a106ebca813160c35a0b16ab#diff-4ed7e979148fc402b0bd376829ccada4R33).  The `scrapy.http.headers.Headers` class has a `.iteritems()` method, while in Python 3 `dict` does not have a `.iteritems()` method.  An update to `FileDownloader` fixed the issue.

Thankfully, I don't think using `.items()` for this segment of `pipeline` code changes the behavior significantly.

davidmcclure:
Ah, sorry @jodizzle, didn't realize `item['headers']` had changed. In `scrapy.http.headers.Headers`, looks like [`items()`](https://github.com/scrapy/scrapy/blob/master/scrapy/http/headers.py#L71) just converts the tuple that comes out of `iteritems()` into a list.

Since `iteritems()` is the method that's actually doing the work, here - and I can't think of a reason to prefer a list to a tuple? - maybe we should revert this and stick with `iteritems()`, just to match the Scrapy API as straightforwardly as possible?

jodizzle:
@davidmcclure Sure, we could revert it if you'd like.  Only situation where `.items()` would be helpful or  necessary is if we ever _wanted_ the headers to be of type `dict`, but I'm not sure when that would be the case.  Using `.iteritems()` is technically more efficient also.  Overall, though, the difference is pretty minor.

PR #166, opened 2017-09-20
Add a filter to avoid undesirable extensions

HarrisonGregg:
This fixes the problem of downloading entire binary files only to discard them.  This occurs because [in `FilterSpider.parse_text`](https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/spiders/__init__.py#L127), any URL that doesn't have an extension in `ALLOWED_FILE_TYPES` is yielded in a request, but [in `FilterSpider.parse_binary`](https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/spiders/__init__.py#L102), any binary response that isn't in `ALLOWED_FILE_TYPES` is discarded.

This fixes the problem by adding a filter that will deny any request whose URL ends in a known extension that we do not want to download.  The filters are checked in downloader middleware, which processes all requests, including redirected ones, so this should cover most cases.

### DIFF
--- a/osp_scraper/filters.py
+++ b/osp_scraper/filters.py
@@ -16,6 +16,28 @@ blacklist_domains = [
     'wikipedia.org',
 ]
 
+# Common file extensions that are not followed if they occur in links
+# modified list from:
+# https://github.com/scrapy/scrapy/blob/dcb279bd6cc85cf1743b548e44b050edba6a2ed8/scrapy/linkextractors/__init__.py
+blacklist_extensions = [
+    # images
+    'mng', 'pct', 'bmp', 'gif', 'jpg', 'jpeg', 'png', 'pst', 'psp', 'tif',
+    'tiff', 'ai', 'drw', 'dxf', 'eps', 'ps', 'svg',
+
+    # audio
+    'mp3', 'wma', 'ogg', 'wav', 'ra', 'aac', 'mid', 'au', 'aiff',
+
+    # video
+    '3gp', 'asf', 'asx', 'avi', 'mov', 'mp4', 'mpg', 'qt', 'rm', 'swf', 'wmv',
+    'm4a',
+
+    # office suites
+    'xls', 'xlsx', 'ppt', 'pptx', 'pps', 'ods', 'odg', 'odp',
+
+    # other
+    'css', 'exe', 'bin', 'rss', 'zip', 'rar',
+]
+
 
 def make_filters(seed_urls, max_hops_from_seed):
     """Generate filters for a spider from a list of URLs.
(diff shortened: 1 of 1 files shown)


wearpants:
Hmm, wonder if this should be a whitelist instead? We should also probably
remove ALLOWED_FILE_TYPES if we go this way

On Sep 20, 2017 2:32 AM, "Harrison" <notifications@github.com> wrote:

This fixes the problem of downloading entire binary files only to discard
them. This occurs because in FilterSpider.parse_text
<https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/spiders/__init__.py#L127>,
any URL that doesn't have an extension in ALLOWED_FILE_TYPES is yielded in
a request, but in FilterSpider.parse_binary
<https://github.com/opensyllabus/osp-scraper/blob/master/osp_scraper/spiders/__init__.py#L102>,
any binary response that isn't in ALLOWED_FILE_TYPES is discarded.

This fixes the problem by adding a filter that will deny any request whose
URL ends in a known extension that we do not want to download. The filters
are checked in downloader middleware, which processes all requests,
including redirected ones, so this should cover most cases.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/opensyllabus/osp-scraper/pull/166
Commit Summary

   - Add a filter to avoid undesirable extensions

File Changes

   - *M* osp_scraper/filters.py
   <https://github.com/opensyllabus/osp-scraper/pull/166/files#diff-0> (33)

Patch Links:

   - https://github.com/opensyllabus/osp-scraper/pull/166.patch
   - https://github.com/opensyllabus/osp-scraper/pull/166.diff

—
You are receiving this because you are subscribed to this thread.
<https://github.com/opensyllabus/osp-scraper/pull/166>, or mute the thread

HarrisonGregg:
@wearpants I guess it just depends on whether we'd like to err on the side of downloading a few files we don't want, or potentially missing a few files that we would want.  It seems like our policy so far has been the former, and I don't think there's a good reason to switch now.  This blacklist should cover > 99% of binary files we don't want, and we could add a log message that tells us whenever there's a URL that returns a binary response that isn't in `ALLOWED_FILE_TYPES` so that we can evaluate how much of an issue it is in the future and find more extensions to blacklist.  That would also be easier than having a whitelist and then trying to identify what extensions we might be missing out on.

osp_scraper/filters.py
```diff
@@ -16,6 +16,22 @@
     'wikipedia.org',
 ]
 
+blacklist_extensions = [
```
jodizzle: I think it would be good to add a comment here explaining where this list came from, i.e., Scrapy and/or that SO post.

osp_scraper/filters.py
```diff
@@ -41,7 +57,20 @@ def make_filters(seed_urls, max_hops_from_seed):
         Filter.compile(
             'deny',
             pattern='regex',
-            hostname=blacklist_re
+            hostname=domain_blacklist_re
+        )
+    )
+
+    extention_blacklist_re = r"^.*\.({})$".format(
```
jodizzle: I think this should be `extension_blacklist_re`, not `extention_blacklist_re`.

Issue #176, opened 2018-01-30
Settings to change


jodizzle:
Before running the next big crawl, we should consider changing some settings.  Here are some notes about various settings that may be valuable to us, ordered (roughly) from most relevant/important to least relevant/important:

- [`DOWNLOAD_MAXSIZE`](https://doc.scrapy.org/en/1.4/topics/settings.html#download-maxsize):  Currently at 128 MiB.  This value seems way to large intuitively, and thankfully we have some empirical evidence to back that up: the highest `Content-Length` of any of our labelled syllabi is around 9.5 MiB.  I think we could triple or maybe even just double that value and be totally fine.
- [`USER_AGENT`](https://doc.scrapy.org/en/1.4/topics/settings.html#user-agent): Currently our user agent is the generic one, "Scrapy/1.4.0 (+http://scrapy.org)".  We should consider using something more identifiable.
- [`ROBOTSTXT_OBEY`](https://doc.scrapy.org/en/1.4/topics/settings.html#robotstxt-obey): Currently set to `True`.  Should we obey robots.txt?
- [`DOWNLOAD_TIMEOUT`](https://doc.scrapy.org/en/1.4/topics/settings.html#download-timeout): Currently set to 180 seconds.  Anecdotally, there seem to be a lot of links that just take forever to resolve ([here's an example](https://lsf.hs-magdeburg.de/)).  This could probably be set to just a couple seconds.
- [`AutoThrottle extension`](https://doc.scrapy.org/en/1.4/topics/autothrottle.html): A variety of settings relating to a way to "automatically adjust scrapy to the optimum crawling speed"; currently disabled for us.  I haven't tried it in a project myself, but it seems reputable, and I think it could really help us get the most from sites, rather than just doing a flat `DOWNLOAD_DELAY` of 3 seconds as we currently do.
- [`RETRY_ENABLED`](https://doc.scrapy.org/en/1.4/topics/downloader-middleware.html#retry-enabled):  Currently set to `True`.  Similar to the `DOWNLOAD_TIMEOUT` question, anecdotally, there are a decent number of sites we visit that are just full of busted links, and repeatedly trying to follow those links can really slow things down.  (Related: [`RETRY_TIMES`](https://doc.scrapy.org/en/1.4/topics/downloader-middleware.html#retry-times).)
[...]

PR #177, opened 2018-02-21
Settings update

jodizzle:
A response to #176.  Sets

- `USER_AGENT: "OSPScraper (+https://www.opensyllabusproject.org)"`
- `ROBOTSTXT_OBEY = False`
- `DOWNLOAD_MAXSIZE = 30 * (1024 * 1024)`

In the settings file, and additionally sets `'DOWNLOAD_TIMEOUT': 60` in the `custom_settings` of FilterSpider.

Also removes little bits of code that conditionally ignored robots.txt, since they're redundant with the new global ignore setting.

### DIFF
--- a/osp_scraper/settings/scrapy.py
+++ b/osp_scraper/settings/scrapy.py
@@ -16,14 +16,15 @@ BOT_NAME = 'osp_scraper'
 SPIDER_MODULES = ['osp_scraper.spiders', 'osp_site_scrapers']
 NEWSPIDER_MODULE = 'osp_site_scrapers'
 
-# Crawl responsibly by identifying yourself (and your website) on the user-agent
-#USER_AGENT = 'osp_scraper (+http://www.yourdomain.com)'
+# NOTE: Add version number of 'osp-scraper'?
+USER_AGENT = "OSPScraper (+https://www.opensyllabusproject.org)"
 
-# Obey robots.txt rules
-ROBOTSTXT_OBEY = True
+# Don't follow robots.txt.
+ROBOTSTXT_OBEY = False
 
-# Maximum file size to download
-DOWNLOAD_MAXSIZE = 128 * (1024 * 1024) # 128 MB
+# Maximum file size to download.  This value is currently three times the
+# largest known size of a syllabus in our collection.
+DOWNLOAD_MAXSIZE = 30 * (1024 * 1024) # 30 MiB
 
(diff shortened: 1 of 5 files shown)


osp_scraper/settings/scrapy.py
```diff
@@ -16,14 +16,15 @@
 SPIDER_MODULES = ['osp_scraper.spiders', 'osp_site_scrapers']
 NEWSPIDER_MODULE = 'osp_site_scrapers'
 
-# Crawl responsibly by identifying yourself (and your website) on the user-agent
-#USER_AGENT = 'osp_scraper (+http://www.yourdomain.com)'
+# NOTE: Add version number of 'osp-scraper'?
+USER_AGENT = "OSPScraper (+https://www.opensyllabusproject.org)"
```
wearpants: no + before the url :-)

osp_scraper/spiders/__init__.py
```diff
@@ -68,6 +68,13 @@ class FilterSpider(OSPSpider):
 
     name = "osp_scraper_spider"
 
+    custom_settings = {
+        **OSPSpider.custom_settings,
+        # Set download timeout to 1 minute (scrapy default is 3 minutes), which
+        # is more appropriate for broad crawls.
+        'DOWNLOAD_TIMEOUT': 60,
```
wearpants: why not put this in scrapy.py with otherr settings?

osp_scraper/settings/scrapy.py
jodizzle: Really?  It seems to be a standard for web crawlers.  [Google uses it in at least some of their crawlers](https://support.google.com/webmasters/answer/1061943?hl=en), and [Scrapy uses it in the default User-Agent](https://doc.scrapy.org/en/latest/topics/settings.html#user-agent).

osp_scraper/spiders/__init__.py
jodizzle: Because I want it to be specific to the broad crawls done by FilterSpider.  I'm not sure how much of a difference it would make, but we definitely designed targeted scrapers for some slow sites and I didn't want to risk breaking anything.

The placement is a bit arbitrary though.  Because we have a class hierarchy of spiders, it kinda opens the question of how much we should be using a global settings file, if at all.  For the moment, I think the distinction I made makes sense.

wearpants:
Ok

jodizzle:
Alright, I'm gonna go ahead and merge.

